PR #14, opened 2018-08-06
Srange

demosdemon:


ptmcg:
Purpose of this change?

pyparsing.py
```diff
@@ -4734,7 +4734,7 @@ def locatedExpr(expr):
 _escapedPunc = Word( _bslash, r"\[]-*.$+^?()~ ", exact=2 ).setParseAction(lambda s,l,t:t[0][1])
 _escapedHexChar = Regex(r"\\0?[xX][0-9a-fA-F]+").setParseAction(lambda s,l,t:unichr(int(t[0].lstrip(r'\0x'),16)))
 _escapedOctChar = Regex(r"\\0[0-7]+").setParseAction(lambda s,l,t:unichr(int(t[0][1:],8)))
-_singleChar = _escapedPunc | _escapedHexChar | _escapedOctChar | Word(printables, excludeChars=r'\]', exact=1) | Regex(r"\w", re.UNICODE)
+_singleChar = _escapedPunc | _escapedHexChar | _escapedOctChar | CharsNotIn(r'\]', exact=1)
```
demosdemon: The purpose of this change is twofold:

1. We are achieving _mostly_ the same effect with one atom instead of two. The `Regex` atom is duplicative of the `Word` atom, except the `Word` atom does not include Unicode characters. 
2. I ran into an issue with the `Regex` atom where `\w` did not match characters I expected it to match. I was translating a grammar into pyparsing; it used a character range `'[A-Z\uSOMECODE-\uSOMECODE]'` _(I don't remember what it was at the moment)_ but the two Unicode characters failed to match. 

I know I could have replaced them with `\\xXXXX` but felt hacky.

demosdemon:
The purpose of this change is twofold:

1. We are achieving _mostly_ the same effect with one atom instead of two. The `Regex` atom is duplicative of the `Word` atom, except the `Word` atom does not include Unicode characters. 
2. I ran into an issue with the `Regex` atom where `\w` did not match characters I expected it to match. I was translating a grammar into pyparsing; it used a character range `'[A-Z\uSOMECODE-\uSOMECODE]'` _(I don't remember what it was at the moment)_ but the two Unicode characters failed to match. 

I know I could have replaced them with `\\xXXXX` but felt hacky.

PR #21, opened 2018-08-17
Boolean Search query parser: allows to perform searches with the common boolean search syntax against a text

xecgr:
Boolean Search query parser, based on searchparser, that allows to perform searches with the common boolean search syntax against a text (western + non-western alphabets)

```
#SAMPLE USAGE:
from booleansearchparser import BooleanSearchParser
bsp = BooleanSearchParser()
text = u"wildcards at the begining of a search term "
exprs= [
    u"*cards and term", #True
    u"wild* and term",  #True
    u"not terms",       #True
    u"terms or begin",  #False
]
for expr in exprs:
    print bsp.match(text,expr)

#non-western samples
text = u"안녕하세요, 당신은 어떠세요?"
exprs= [
    u"*신은 and 어떠세요", #True
    u"not 당신은",       #False
    u"당신 or 당",  #False
]
for expr in exprs:
    print bsp.match(text,expr)
```

### DIFF
--- a/examples/booleansearchparser.py
+++ b/examples/booleansearchparser.py
@@ -21,11 +21,6 @@ Requirements:
 * Pyparsing
 
 
-CONTRIBUTORS:
-- Guiem Bosch
-- Francesc Garcia
-
-
 SAMPLE USAGE:
 from booleansearchparser import BooleanSearchParser
 from __future__ import print_function
@@ -80,6 +75,8 @@ CONTRIBUTORS:
 - Steven Mooij
 - Rudolph Froger
 - Paul McGuire
+- Guiem Bosch
+- Francesc Garcia
 
 TODO:
 - add more docs
@@ -89,7 +86,6 @@ TODO:
 """
 from __future__ import print_function
 from pyparsing import Word, alphanums, Keyword, Group, Combine, Forward, Suppress, Optional, OneOrMore, oneOf
-from sets import Set
 import re
 import string
 alphabet_ranges = [
@@ -152,7 +148,7 @@ class BooleanSearchParser:
         )
         #suport for non-wester alphabets
         for r in alphabet_ranges:
-            alphabet += u''.join(unichr(c) for c in xrange(*r) if not unichr(c).isspace())
+            alphabet += u''.join(chr(c) for c in range(*r) if not chr(c).isspace())
         
         operatorWord = Group(
             Word(alphabet+'*')
@@ -213,7 +209,7 @@ class BooleanSearchParser:
         function GetQuoted to only return the subset of ID's that contain the
         literal string.
         """
-        #r = Set()
+        #r = set()
         r = False
         search_terms = []
         for item in argument:
@@ -264,15 +260,15 @@ class BooleanSearchParser:
     
     """
     def GetKeyword(self, name, value):
-        return Set()
+        return set()
 
     def GetBetween(self, min, max):
         print (min,max)
-        return Set()
+        return set()
     """
     
     def GetQuotes(self, search_string, tmp_result):
-        return self.text.count(search_string) > 0
+        return search_string in self.text
     
 
     def GetNot(self, not_set):
@@ -286,7 +282,7 @@ class BooleanSearchParser:
         >>> string.punctuation
         '!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~'
         """
-        #it will keep @, # and _
+        #it will keep @, # and 
         #usernames and hashtags can contain dots, so a double check is done
         r = re.compile(r'[\s{}]+'.format(re.escape('!"$%&\'()*+,-/:;<=>?[\\]^`{|}~')))
         _words  = r.split(text)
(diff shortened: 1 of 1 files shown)


ptmcg:
Thanks for submitting this - I'll review for inclusion in the examples in the next release (I held back a number of other items from 2.2.1 to focus on the `DeprecationWarning` issue and the first release from GitHub).

ptmcg:
I've reviewed this code, and you have added some nice enhancements over and above the search query parser, such as the use of wildcards, and non-Western alphabets. If you don't mind, I will rework the non-Western alphabets to use the new pyparsing_unicode ranges (this actually fixes a subtle bug in your current code, in that the last character of each range is left out because of the inclusive-start-exclusive-end nature of range/xrange).

ptmcg:
This might also require some fixes to handle "a and b and c" queries, and "a and b or c and d or e".

examples/booleansearchparser.py
```diff
@@ -0,0 +1,486 @@
...
+
+#non-western samples
+text = u"안녕하세요, 당신은 어떠세요?"
+exprs= [
+    u"*신은 and 어떠세요", #True
+    u"not 당신은",       #False
+    u"당신 or 당",  #False
+]
+for expr in exprs:
+    print bsp.match(text,expr)
```
cthoyt: I would suggest using `from __future__ import print_function` and changing this over to be Python 2/3 compatible

xecgr: Sure, shall I add the changes o can you add it yourself?

cthoyt: I'm sorry but I don't think I can edit your pull request.

examples/booleansearchparser.py
```diff
@@ -0,0 +1,486 @@
...
+        """
+        operatorOr = Forward()
+        
+        alphabet  = ( 
+            u'*'+
+            alphanums
+        )
+        #suport for non-wester alphabets
+        for r in alphabet_ranges:
+            alphabet += u''.join(unichr(c) for c in xrange(*r) if not unichr(c).isspace())
```
sirosen: I saw this PR and thought I should offer: I tried something similar to this and found that it was easier to define my alphabet using a Regex expression which _excludes_ special characters. I did this instead of using pyparsing.Word so that I could use a Regex with `+` -- in my tests this was more efficient than any of the other approaches I tried.

```
# ESCAPE = \ , ESCAPED_SPECIAL = ESCAPE + Word(OPERATOR | ESCAPE, exact=1)
# not included here for brevity
ATOM = Combine(
    OneOrMore(
        ESCAPED_SPECIAL |
        Regex(r"""[^'":\\()/+\-=&|><!{}\[\]^~? ]+""") |
        ESCAPE
    )
)
```

klahnakoski:
@xecgr, @ptmcg    What is happening with this PR?  Should it be closed for lack of activity?

ptmcg:
Ultimately, this PR should be an update to the existing search query parser example, instead of a clone of it, with some enhancements added. I like the changes, but let's just do them against the existing example.

examples/booleansearchparser.py
xecgr: I've added your suggestion : https://github.com/xecgr/pyparsing/commit/f07b5776df171d3572e616644b48748dee320c18

xecgr:
sorry for the delay, obviosly I forgot to answer some questions...
I've updated the code adding the @cthoyt suggestion. The @sirosen proposal differs from mine, because he uses a pyparsing expression instead of python code (the option that I chose). What approach would be better?

ptmcg:
I'll review again with latest changes today, thanks for hanging in there!

examples/booleansearchparser.py
```diff
@@ -0,0 +1,488 @@
...
+    def GetKeyword(self, name, value):
+        return Set()
+
+    def GetBetween(self, min, max):
+        print (min,max)
+        return Set()
+    """
+    
+    def GetQuotes(self, search_string, tmp_result):
+        return self.text.count(search_string) > 0
```
ptmcg: Better written as:

    return search_string in self.text

count() will scan the entire string, when it is only necessary to scan to the first occurrence.

examples/booleansearchparser.py
```diff
@@ -0,0 +1,488 @@
...
+                "10", 
+                "12"
+            ]
+        }
+
+        
+        all_ok = True
+        for text,matches in texts_matcheswith.iteritems():
+            _matches = []
+            for _id,expr in exprs.iteritems():
```
ptmcg: `iteritems()` is Py2-only, change to using `items()`

xecgr:
@ptmcg I've commited requested changes. Could you check them?

ptmcg:
Thanks for being so patient! Many things happened with pyparsing this year, and I'm thankful for your submission!
-- Paul

ptmcg:
I tried adding your non-western examples from the doc string to the test cases, but the tests failed. Could you please troubleshoot and submit a new PR?

xecgr:
Could you share the error screenhost?
I've tried them and they are fine
Screen Shot 2019-08-14 at 12 58 59

ptmcg:
Sorry, I meant, could you please add them to the tests in ParserTest?

xecgr:
I've added test to ParserTest. It raised an incompatibility py2/py3 solved in commit 5988ecdbb4ac9c0b6f69df8e6e04bdf6dd4b8f75
Let me know if everything works fine

ptmcg:
Ok, I found the issue, I had left out other matching expressions texts_matcheswith. Pushed updates, all looks good now.

PR #46, opened 2018-11-20
Fixing distribution issues

### LINKED ISSUE #44: Distribute tests in pypi tarball
Hello in various distributions we try to make sure the packages we in the end ship to users are really working and for that we execute upstream unittests to ensure some validity. Esp. with python that is not compiled it is quite good idea to check that with distribution patching in python we didn't break something.

As such could you please include in Manifest.in the files executed by the tox/travis for us to validate against?

mcepl:
Originally I wanted to fix only ``MANIFEST.in`` file, but while doing that I have discovered that the package has practically unrebuildable documentation on openSUSE, because we don't have epydoc in the main repository any more.

So, I have converted documentation to Sphinx, which feels like a default documentation generator these days.

It is still WIP, because there are still bugs in the documentation strings of many functions, but the documentation builds with ``cd docs ; make html`` (or any other format supported by Sphinx).

Yes, fixes #44, but not only that.


mcepl:
I think could be merged although work on reformatting documentation strings could continue.

ptmcg:
I have just about finished cleaning up the remaining Sphinx errors

PR #54, opened 2018-12-21
Remove distutils fallback and outdated comment

jdufresne:
setuptools includes a vendored version of pyparsing (and other
dependencies). They are not install through traditional tools.
Therefore, distutils is not required as fallback to facilitate
setuptools.

https://github.com/pypa/setuptools/blob/v40.6.3/setuptools/_vendor/pyparsing.py

Also remove the unused imports.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,9 @@ Version 2.3.1 -
   for running tests that do additional validation or processing on the parsed
   results.
 
+- Removed distutils fallback in setup.py. If installing the package fails,
+  please update to the latest version of setuptools.
+
 
 Version 2.3.0 - October, 2018
 -----------------------------
--- a/setup.py
+++ b/setup.py
@@ -2,18 +2,9 @@
 
 """Setup script for the pyparsing module distribution."""
 
-# Setuptools depends on pyparsing (via packaging) as of version 34, so allow
-# installing without it to avoid bootstrap problems.
-try:
-    from setuptools import setup
-except ImportError:
-    from distutils.core import setup
-
-import sys
-import os
-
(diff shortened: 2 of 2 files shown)


ptmcg:
Thanks for this PR, looks great. I try to maintain some level of backward compatibility with existing platforms and deployments, can you comment on whether these changes will leave those on older platforms high and dry? Or is it simply a matter of "if this doesn't work for you, you should update to the latest setuptools."? (which we could add to the pyparsing CHANGES file)

jdufresne:
> can you comment on whether these changes will leave those on older platforms high and dry?

For this to fail, it would mean the user tried to install the package with the command `python setup.py install` (not through pip) and the Python installation doesn't have setuptools installed. In my experience, not having setuptools installed is extremely unusual and not something I've personally encountered. The Python Windows installation includes setuptools and most (if not all) Linux distros install setuptools by default. Many other packages I use or contribute to intentionally omit the distutils fallback. But the concern is very fair and worth considering.

I'll note that the Python documentation itself recommends against using distutils directly:

https://docs.python.org/3/library/distutils.html

> Most Python users will not want to use this module directly, but instead use the cross-version tools maintained by the Python Packaging Authority. In particular, setuptools is an enhanced alternative to distutils that provides ...

---

> Or is it simply a matter of "if this doesn't work for you, you should update to the latest setuptools."? (which we could add to the pyparsing CHANGES file)

Good idea! I've added a line to CHANGES in the latest revision.

Thanks!

ptmcg:
Please comment on https://github.com/pyparsing/pyparsing/pull/133, which proposes undoing this change.

PR #71, opened 2019-02-16
Descriptive names for Forward expressions

eswald:
Resolves the infinite recursion potential by setting a temporary name during resolution.


ptmcg:
Thanks for submitting, I wasn't aware this was a problem. I'll look at it in the next few days.

PR #72, opened 2019-02-16
Support for simple_unit_tests under Python 2

eswald:
This might not be too desirable with the end of life right around the corner, but the incompatibilities weren't too tricky to deal with for the purposes of simply running the suite.  Granted, the traceback sent to `explain` is probably the wrong one, and the `subTest` shim isn't fully functional, but at least the file passes.


ptmcg:
Thanks for submitting, I'll look at it in the next few days.

PR #75, opened 2019-03-01
Fix cases where an indentedBlock partially parsing will cause all other correct indentedBlocks to fail

### LINKED ISSUE #74: indentedBlock not clearing the indent stack when it partially matches, and then fails
First of all, thanks for this amazingly powerful and expressive library. I'm so glad to have found it.

I want to parse a language that uses indentation semantically, and thus I am using the `indentedBlock` function. However, I want to use it in combination with `scanString()`, because not all text in my input string belongs to this language. 

However, in doing this I have noticed a bug in the `indentedBlock`. If you try to parse a string that *does* include an indented block, but *doesn't* completely parse, the indent stack will not be reverted to how it was, and thus it will fail to parse all correct statements.

To demonstrate this, I've made a simple test case: https://github.com/TMiguelT/PyparsingIndent/blob/master/indent.py. The comments explain that, if the `indentedBlock` expression matches, but then the rest of the parsing fails, then the parser will fail to match anything from then on.

TMiguelT:
Fixes #74, as described in that issue

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" project-jdk-name="Python 3.6 (pyparsing)" project-jdk-type="Python SDK" />
+  <component name="ProjectRootManager" version="2" project-jdk-name="Python 3.6.3 (/usr/bin/python3.6)" project-jdk-type="Python SDK" />
 </project>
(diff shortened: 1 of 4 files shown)


TMiguelT:
Oh, and I added tests!

TMiguelT:
Thanks!

TMiguelT:
Is it likely there will be a release soon, incorporating this PR? I only ask because I want to publish something to pypi that depends upon a fixed pyparsing.

ptmcg:
I will try to get it out this week. My misstep with the 2.3.1 release (fixed a bug which changed the API in a subtle but pervasive way) is slowing everything down.

PR #83, opened 2019-04-06
snmp_api.h missing from manifest

bdragon28:
Add examples/*.h to manifest so snmp_api.h gets packaged with the rest of the test files.
Noticed when attempting to run "python3 setup.py test" against 2.3.1 tarball.


bdragon28:
there's still something a bit off about this. The test still had issues until I copied snmp_api.h.

Is running `python3 setup.py test` currently unsupported?

ptmcg:
I'm behind the curve on setuptools. Looked up "python setup.py test", and am making changes so that this will run the basic unitTests.py.

PR #98, opened 2019-06-11
Support for omitted integer part of the float number

ansobolev:
When writing float numbers, integer part can be omitted, in which case it defaults to 0. Thus, `.123456` and `-.123456` are valid floats, equalling to 0.123456 and -0.123456, respectively. However, `pyparsing` does not support such notation. This pull request changes the quantifier of the `pyparsing_common.real` integer part regex token, adding support for such numbers.

### DIFF
--- a/pyparsing.py
+++ b/pyparsing.py
@@ -6153,7 +6153,7 @@ class pyparsing_common:
     """mixed integer of the form 'integer - fraction', with optional leading integer, returns float"""
     mixed_integer.addParseAction(sum)
 
-    real = Regex(r'[+-]?\d+\.\d*').setName("real number").setParseAction(convertToFloat)
+    real = Regex(r'[+-]?\d*\.\d*').setName("real number").setParseAction(convertToFloat)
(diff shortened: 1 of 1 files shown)


ptmcg:
I actually did this intentionally, harkening back to my early learning days with the Pascal language, which required a leading '0' before the decimal point. But Python accepts a numeric string without a leading digit, so  as your change is consistent with Python, it makes more sense to accept this less-rigorous format. Should this change also be applicable to floating point numbers using scientific notation, which is parsed using the `pyparsing_common.sci_real` expression?

ansobolev:
That's a good question. Python apparently accepts numbers such as `-.123e3`; however, this is a highly non-conventional form of scientific notation. I have never encountered numbers written in such a way (this is only anecdotal evidence though). But if the numbers _can_ be written like that, why not let them be parsed?

ptmcg:
I'll also add new tests to the pyparsing unit tests for these formats - it turns out to be a little more complicated than the simple change in your PR, but nothing too unsavory.

PR #113, opened 2019-08-12
Update example protobuf to proto3 syntax

loonydev:
Hey, tried to use this example on proto3 syntax and got a problem. Hope this fix will be useful for somebody.

Regarding to: https://developers.google.com/protocol-buffers/docs/proto3

### DIFF
--- a/examples/protobuf_parser.py
+++ b/examples/protobuf_parser.py
@@ -15,7 +15,7 @@ integer = Regex(r"[+-]?\d+")
 LBRACE,RBRACE,LBRACK,RBRACK,LPAR,RPAR,EQ,SEMI = map(Suppress,"{}[]()=;")
 
 kwds = """message required optional repeated enum extensions extends extend
-          to package service rpc returns true false option import"""
+          to package service rpc returns true false option import syntax"""
 for kw in kwds.split():
     exec("{0}_ = Keyword('{1}')".format(kw.upper(), kw))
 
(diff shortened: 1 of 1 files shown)


examples/protobuf_parser.py
```diff
@@ -50,6 +50,8 @@
 # serviceDefn ::= 'service' ident '{' methodDefn* '}'
 serviceDefn = SERVICE_ - ident("serviceName") + LBRACE + ZeroOrMore(Group(methodDefn)) + RBRACE
 
+syntaxDefn = SYNTAX_ + EQ - quotedString("syntaxString") + SEMI
```
ptmcg: I'd rather not call this "syntaxString" in favor of just "syntax".

examples/protobuf_parser.py
```diff
@@ -27,8 +27,8 @@
                     fixed32 fixed64 sfixed32 sfixed64 bool string bytes""") | ident
 rvalue = integer | TRUE_ | FALSE_ | ident
 fieldDirective = LBRACK + Group(ident + EQ + rvalue) + RBRACK
-fieldDefn = (( REQUIRED_ | OPTIONAL_ | REPEATED_ )("fieldQualifier") -
-              typespec("typespec") + ident("ident") + EQ + integer("fieldint") + ZeroOrMore(fieldDirective) + SEMI)
+fieldDefnPrefix =  REQUIRED_ | OPTIONAL_ | REPEATED_
+fieldDefn = (Optional(fieldDefnPrefix) + typespec("typespec") + ident("ident") + EQ + integer("fieldint") + ZeroOrMore(fieldDirective) + SEMI)
```
ptmcg: Please restore "fieldQualifier" as a results name on `fieldDefnPrefix`, something like:

    fieldDefn = (Optional(fieldDefnPrefix("fieldQualifier")) + typespec("typespec") + etc.

examples/protobuf_parser.py
```diff
@@ -97,4 +99,15 @@
   repeated Person person = 1;
 }"""
 
-parser.runTests([test1, test2])
```
ptmcg: And added a test too, thanks!

ptmcg:
Thanks so much for submitting - please see review (minor) review comments.

examples/protobuf_parser.py
ptmcg: Looking at the spec further, this might be better named "fieldRule". In proto3, I don't see "required" or "optional", but there are references to "singular" and "repeated". Is there an actual BNF somewhere?

loonydev: As I know "required" are optional, but still in syntax.

examples/protobuf_parser.py
loonydev: Nothing to do, you missed my changes at L113
```suggestion
parser.runTests([test1, test2, test3])
```

PR #117, opened 2019-08-14
parseAll implementation by grammar modification

cngkaygusuz:
Hi Paul,

The reason I'm proposing this is the inconsistency in the `ParseException.explain` method, illustrated as follows:

First example, where `parseAll=True`:

```python
expr = Word('012') + Word('abc')
try:
    expr.parseString('00aaf', parseAll=True)
except ParseException as pe:
    print(ParseException.explain(pe))
```

prints

```
00aaf
    ^
ParseException: Expected end of text, found 'f'  (at char 4), (line:1, col:5)
pyparsing.And - {Empty StringEnd}
pyparsing.StringEnd - StringEnd
```

The second example, where the same effect is accomplished with `StringEnd`

```python
expr = Word('012') + Word('abc') + StringEnd()

try:
    expr.parseString('00aaf')
except ParseException as pe:
    print(ParseException.explain(pe))
```

prints

```
00aaf
    ^
ParseException: Expected end of text, found 'f'  (at char 4), (line:1, col:5)
pyparsing.And - {W:(012) W:(abc) StringEnd}
pyparsing.StringEnd - StringEnd
```

Observe the difference in output on line starting with `pyparsing.And`: the first approach parses the leftover string with a grammar defined as `Empty() + StringEnd()`. Upon receiving the exception, the grammar trace reports a grammar that does not have to do with the original grammar description, whereas the second approach clearly indicates a `StringEnd()` has been appended to the original grammar, which I believe is the better output format in an obvious way. 

With this reasoning, I am modifying the implementation of `parseString` so that, if `parseAll==True`, the parser element constructs a new parser by appending `StringEnd()` to itself, and delegating the parsing action to the new parser's `parseString` method.

Please note that this work is not yet ready to be merged, as you can see there aren't any tests added and I am not confident this modification won't break anything. If you give it a go, I shall examine other parser entry to ensure uniform behavior, and make sure it does not break anything.

### DIFF
--- a/pyparsing.py
+++ b/pyparsing.py
@@ -1752,21 +1752,23 @@ class ParserElement(object):
         ...
         pyparsing.ParseException: Expected end of text, found 'b'  (at char 5), (line:1, col:6)
         """
-        
+        if parseAll:
+            parser = self + StringEnd()
+            return parser.parseString(instring)
+
         ParserElement.resetCache()
+
         if not self.streamlined:
             self.streamline()
-            # ~ self.saveAsList = True
+
         for e in self.ignoreExprs:
             e.streamline()
+
         if not self.keepTabs:
             instring = instring.expandtabs()
+
         try:
             loc, tokens = self._parse(instring, 0)
-            if parseAll:
-                loc = self.preParse(instring, loc)
-                se = Empty() + StringEnd()
-                se._parse(instring, loc)
         except ParseBaseException as exc:
             if ParserElement.verbose_stacktrace:
                 raise
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,10 +1,10 @@
 # This file contains the configuration for Scrutinizer-CI, a tool we use for software quality purposes.
 build:
   environment:
-    python: 3.6.3
+    python: 3.7.2
 
     variables:
-      PYTHON_VERSIONS: 'jython-2.7.1 pypy2.7-5.10.0 pypy3.5-5.10.1 2.7.15 3.3.6 3.4.5 3.5.6'
+      PYTHON_VERSIONS: 'pypy3.5-6.0.0 3.5.6 3.6.3'
 
   dependencies:
     override:
--- a/scrutinizer-pyenv.sh
+++ b/scrutinizer-pyenv.sh
@@ -8,4 +8,4 @@ git fetch --tags
 git checkout v1.2.7
 popd
 
-echo $PYTHON_VERSIONS | xargs -n1 pyenv install
+echo $PYTHON_VERSIONS | xargs -n1 pyenv install -s
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist=pypy, pypy3, py27, py33, py34, py35, py36
+envlist=pypy3, py35, py36, py37
 
 [testenv]
 deps=-rrequirements-dev.txt


ptmcg:
Going on vacation, will not get to look at this for a week or so. I'm a little surprised by this, but nothing should surprise me any more, really.  Thanks for the notes.

pyparsing.py
```diff
@@ -1752,21 +1752,23 @@ def parseString(self, instring, parseAll=False):
         ...
         pyparsing.ParseException: Expected end of text, found 'b'  (at char 5), (line:1, col:6)
         """
-        
+        if parseAll:
```
ptmcg: This change is necessary to pass unit tests:

        if parseAll:
            str_end = StringEnd()
            for expr in self.ignoreExprs:
                str_end.ignore(expr)
            return (self + str_end).parseString(instring)

Also, I think this PR should contain only this change and the other changes you've proposed to pyparsing.py - the other changed files should be part of a different PR.

ptmcg:
I was about to accept this PR, until I added unit tests to demonstrate the new behavior. Your example looks good when used with a small test expression, but many parsers that will use `parseAll=True` will be very complex, and have long str representations. 

Instead of your sample expression:

    expr = Word('012') + Word('abc') + StringEnd()

try this more extreme example (simulating a complex parser):

    anything_but_an_f = OneOrMore(MatchFirst([Literal(c) for c in printables if c != 'f']))
    expr = Word('012') + anything_but_an_f + StringEnd()

In my unit test, I got an `explain()` string that looks like:

    00aaf
        ^
    ParseException: Expected end of text, found 'f'  (at char 4), (line:1, col:5)
    __main__.ParseAllTest
    pyparsing.And - {W:(012) {{"0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" | "a" | "b" | "c" | "d" | "e" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z" | "A" | "B" | "C" | "D" | "E" | "F" | "G" | "H" | "I" | "J" | "K" | "L" | "M" | "N" | "O" | "P" | "Q" | "R" | "S" | "T" | "U" | "V" | "W" | "X" | "Y" | "Z" | "!" | """ | "#" | "$" | "%" | "&" | "'" | "(" | ")" | "*" | "+" | "," | "-" | "." | "/" | ":" | ";" | "<" | "=" | ">" | "?" | "@" | "[" | "\" | "]" | "^" | "_" | "`" | "{" | "|" | "}" | "~"}}... StringEnd}
    pyparsing.StringEnd - StringEnd

In illustrating that there was more text in the input string than was expected by the parser, I think the message `"Expected end of text, found 'f'  (at char 4), (line:1, col:5)"` and highlighting the error location is sufficient without repeating the original expression.

I do agree that the `Empty()` expression confuses things here. In diagnosing the problem with the `ignoreExprs` not being copied to the added `StringEnd()`, I wondered why the added `Empty()` wasn't having this problem. It turns out that the `Empty()` wasn't what made things work, it was the previous line's call to `self.preParse()`. I think I had inserted the `Empty()` as my first attempt at whitespace and ignorable skipping before testing for `StringEnd()`, it didn't work, I added the `preParse` call which *did* work, and the `Empty()` just got left in. If we revert the code to its original form, and just remove the `Empty()`, we get an explain text that reads:

    00aaf
        ^
    ParseException: Expected end of text, found 'f'  (at char 4), (line:1, col:5)
    __main__.ParseAllTest
    pyparsing.StringEnd - StringEnd
    <<<< End of test ParseAllTest

I don't think this loses anything of diagnostic value, when we are simply trying to indicate that there was more text beyond what was expected by the parser.

So I think the best way to address your issue is to just change:

                se = Empty() + StringEnd()
to:

                se = StringEnd()

Your other code cleanups in pyparsing.py also look fine. As I mentioned in my review comment, I think the other file changes should be in a separate PR.

pyparsing.py
ptmcg: No, this is wrong, see more recent comments on the PR.

cngkaygusuz:
Well, my initial goal was not to merge this as is right now, but to invoke a response just like you have posted. This PR didn't even clear the tests already present in the code, but I submitted this nevertheless as I find communicating in the presence of concrete code facilitates a much better conversation. There is quite a lot of history in this codebase which the contributors should have a healthy grasp of, instead of going through the code and explaining things, I think it is best to unearth them when they matter, just like we did here.

I pretty much agree with everything here and will do the necessary modifications and submissions.

PR #126, opened 2019-09-20
BigQuery View parse fails on IGNORE NULLS

smedberg:
BigQuery view SQL parser was failing on IGNORE NULL statements like:

```
SELECT
  DISTINCT FIRST_VALUE(x IGNORE NULLS) OVER (PARTITION BY y)
FROM z
```

E.g. see doc at https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#first_value


ptmcg:
Is there a BNF somewhere that fully describes this SQL dialect? If so, you should be able to validate your pyparsing code against it to see if there are other omissions like this.

smedberg:
@ptmcg Sorry for never getting back to you.  I'm not aware of a BNF that describe this dialect, but that doesn't mean there isn't one :).  I've spent a little time looking, but I haven't contacted AWS support.

PR #130, opened 2019-10-08
refactor tests

klahnakoski:


tests/test_unit.py
```diff
@@ -7,23 +7,33 @@
...
+from examples.jsonParser import jsonObject
+from pyparsing import ParseException
+
+ppt = pp.pyparsing_test
+TestParseResultsAsserts = ppt.TestParseResultsAsserts
+
+try:
+    from tests.json_parser_tests import test1, test2, test3, test4, test5
+except Exception as e:
+    from json_parser_tests import test1, test2, test3, test4, test5
```
klahnakoski: @ptmcg   for some reason the command line and pyCharm use different import rules (absolute and relative respectively) . Until I find out how to fix that, I import both.

pyparsing.py
```diff
@@ -6719,25 +6719,34 @@ def __init__(self):
             self._save_context = {}
 
         def __enter__(self):
+            self.save()
```
klahnakoski: I renamed these so they were more descriptive in the test suite that used them

pyparsing.py
```diff
@@ -6719,25 +6719,34 @@ def __init__(self):
...
+            self.restore()
+
+        def save(self):
             self._save_context['default_whitespace'] = ParserElement.DEFAULT_WHITE_CHARS
             self._save_context['default_keyword_chars'] = Keyword.DEFAULT_KEYWORD_CHARS
             self._save_context['literal_string_class'] = ParserElement._literalStringClass
             self._save_context['packrat_enabled'] = ParserElement._packratEnabled
+            self._save_context['packrat_parse'] = ParserElement._parse
             self._save_context['__diag__'] = {name: getattr(__diag__, name) for name in __diag__._all_names}
+            self._save_context['__compat__'] = {"collect_all_And_tokens": __compat__.collect_all_And_tokens}
```
klahnakoski: These are the two 'global' states that were missed.

tests/test_examples.py
```diff
@@ -0,0 +1,30 @@
+from importlib import import_module
+import unittest
+
+
+class TestExamples(unittest.TestCase):
+
+    def _run(self, name):
+        import_module('examples.'+name)
```
klahnakoski: This class will allow us to run the examples as part of the test suite.  I have seen these particular examples run in travis.yml

.travis.yml
```diff
@@ -17,15 +17,7 @@ install:
...
- - python unitTests.py
- - python simple_unit_tests.py
- - PYTHONPATH=. python examples/numerics.py
- - PYTHONPATH=. python examples/TAP.py
- - PYTHONPATH=. python examples/romanNumerals.py
- - PYTHONPATH=. python examples/sexpParser.py
- - PYTHONPATH=. python examples/oc.py
- - PYTHONPATH=. python examples/delta_time.py
- - PYTHONPATH=. python examples/eval_arith.py
+ - python -m unittest discover tests
```
klahnakoski: Simplified running of tests.  There is a new test class that will run examples as test for us.  I assume an example will fail if it can not parse.

tests/test_simple_unit.py
```diff
@@ -11,15 +11,17 @@
...
+TestParseResultsAsserts = ppt.TestParseResultsAsserts
 
 # Test spec data class for specifying simple pyparsing test cases
 PpTestSpec = namedtuple("PpTestSpec", "desc expr text parse_fn "
                                       "expected_list expected_dict expected_fail_locn")
 PpTestSpec.__new__.__defaults__ = ('', pp.Empty(), '', 'parseString', None, None, None)
 
 
-class PyparsingExpressionTestCase(ppt.ParseResultsAsserts, unittest.TestCase):
+class PyparsingExpressionTestCase(TestParseResultsAsserts):
```
klahnakoski: since `TestParseResultsAsserts` extends `TestCase` we do not need the multiple inheritance (plus it was confusing pyCharm's simple test discovery logic).

tests/test_unit.py
```diff
@@ -61,51 +71,21 @@ def __exit__(self, *args):
...
-
-                finally:
-                    print("<<<< End of test", str(self))
-                    print()
-
-        except Exception as exc:
-            if BUFFER_OUTPUT:
-                print()
-                print(buffered_stdout.getvalue())
-            raise
```
klahnakoski: As discussed here https://github.com/orgs/pyparsing/teams/pyparsing-dev/discussions/5?from_comment=11#discussion-5-comment-11 unittest does this.

tests/test_unit.py
```diff
@@ -61,51 +71,21 @@ def __exit__(self, *args):
...
 
+class Test1_PyparsingTestInit(TestCase):
```
klahnakoski: The "Test" prefix is required so pyCharm can see these as tests.  The `Test1`, `Test2`, ...  pattern ensures these test classes are run in order.

tests/test_unit.py
```diff
@@ -162,8 +142,7 @@ def runTest(self):
         finally:
             pp.ParserElement.setDefaultWhitespaceChars(prev_default_whitespace_chars)
 
-class UpdateDefaultWhitespace2Test(ParseTestCase):
-    def runTest(self):
+    def testUpdateDefaultWhitespace2(self):
```
klahnakoski: All test classes are converted to test methods, and renamed to start with "test"

tests/test_unit.py
```diff
@@ -1719,17 +1667,7 @@ class AddOp(BinOp):
...
 
 
-class PickleTest_Greeting():
-    def __init__(self, toks):
-        self.salutation = toks[0]
-        self.greetee = toks[1]
-
-    def __repr__(self):
-        return "%s: {%s}" % (self.__class__.__name__,
-            ', '.join('%r: %r' % (k, getattr(self, k)) for k in sorted(self.__dict__)))
```
klahnakoski: This class has been moved to the bottom of the file

tests/test_unit.py
```diff
@@ -2433,13 +2357,7 @@ def __str__(self):
                          "Failed to parse using variable length parse actions "
                          "using class constructors as parse actions")
 
-class EnablePackratParsing(TestCase):
-    def runTest(self):
-        from pyparsing import ParserElement
-        ParserElement.enablePackrat()
```
klahnakoski: moved to bottom of file

tests/test_unit.py
```diff
@@ -2933,14 +2839,7 @@ def runTest4(self):
             """)
 
 
-    def runTest(self):
-        self.runTest1()
-        self.runTest2()
-        self.runTest3()
-        self.runTest4()
```
klahnakoski: each test was given its own, more descriptive, name.  Since they are methods, they are now tests (look above)

tests/test_unit.py
```diff
@@ -4433,38 +4291,37 @@ def key_parse_action(toks):
...
-        """
-        A valid statement is the word "block:", followed by an indent, followed by the letter A only, or another block
-        """
-        stack = [1]
-        block = pp.Forward()
-        body = pp.indentedBlock(pp.Literal('A') ^ block, indentStack=stack, indent=True)
-        block <<= pp.Literal('block:') + body
-        return block
+    def testIndentedBlockScan(self):
+        def get_parser():
```
klahnakoski: This could be moved to the bottom of the file, but I thought keeping it local to the tests that use it would be good.  So it is an inner method.

tests/test_unit.py
```diff
@@ -5148,61 +4990,33 @@ def getNameTester(s, l, t):
             self.assertEqual(len(stmt[1, ...]('tests').parseString('test test').tests), 2, "multipled(3) failure with setResultsName")
             self.assertEqual(len(stmt[2, ...]('tests').parseString('test test').tests), 2, "multipled(3) failure with setResultsName")
 
-def makeTestSuite():
```
klahnakoski: no longer required; unittest will effectively assemble the same.

tests/test_unit.py
```diff
@@ -5148,61 +4990,33 @@ def getNameTester(s, l, t):
...
-    suite.addTest(PyparsingTestInit())
-
-    test_case_classes = ParseTestCase.__subclasses__()
-    # put classes in order as they are listed in the source code
-    test_case_classes.sort(key=lambda cls: inspect.getsourcelines(cls)[1])
 
-    test_case_classes.remove(PyparsingTestInit)
-    # test_case_classes.remove(ParseASMLTest)
-    if IRON_PYTHON_ENV:
-        test_case_classes.remove(OriginalTextForTest)
```
klahnakoski: I am not sure why this test was excluded.  If it still needs to be then we can annotate the method

        @skipIf(IRON_PYTHON_ENV, "not expected to work in iron python")

tests/test_unit.py
```diff
@@ -5148,61 +4990,33 @@ def getNameTester(s, l, t):
...
-
-    if TEST_USING_PACKRAT:
-        # retest using packrat parsing (disable those tests that aren't compatible)
-        suite.addTest(EnablePackratParsing())
+class PickleTest_Greeting():
+    def __init__(self, toks):
+        self.salutation = toks[0]
+        self.greetee = toks[1]
 
-        unpackrattables = [PyparsingTestInit, EnablePackratParsing, RepeaterTest,]
```
klahnakoski: The `RepeaterTest` (now called `testRepeater`) already checks if the packrat is on, and exits early if so.  No need to skip.

tests/test_unit.py
```diff
@@ -5148,61 +4990,33 @@ def getNameTester(s, l, t):
...
+        Test2_WithoutPackrat.suite_context = ppt.reset_pyparsing_context()
+        Test2_WithoutPackrat.suite_context.save()
 
-# runnable from setup.py using "python setup.py test -s unitTests.suite"
-suite = makeTestSuite()
+Test4_WithPackrat = type(
+    "Test4_WithPackrat",
+    (Test2_WithoutPackrat,),
+    {}
+)
```
klahnakoski: This declares a new class, that inherits from `(Test2_WithoutPackrat,)` and adds no new attributes.  Now we have two classes that do the same thing, but with packrat mode being different.

tests/test_unit.py
ptmcg: In PyCharm, edit your run configuration for 'UnitTests' to set the parent directory as the working directory. This should then let you fix on absolute imports.

tests/test_unit.py
ptmcg: At one time, the behavior in this test did not run correctly on IronPython. I don't know if this is still an issue or not.

tests/test_unit.py
klahnakoski: fixed with https://github.com/pyparsing/pyparsing/pull/130/commits/dc70de952da931d655502c41a0db80fbd35e9571

tests/test_unit.py
klahnakoski: I suggest we do not skip the test.  If/when it fails in IronPython, then we have something to fix

ptmcg:
@klahnakoski - can you please resolve the outstanding conflicts, then I can merge you changes.

klahnakoski:
@ptmcg this PR consists of multiple commits, and a merge from master.  Some module owners like their PRs to be compressed to a single commit, and rebased onto master.  I am from another camp that lets the history be what it is; letting the PR represent the net-effect.

You have a preference?

ptmcg:
I've been doing squash/merge on PRs - I think that otherwise the commit history just gets flooded with many small commits, instead of just sticking to the significant ones.

klahnakoski:
squashed and rebased

PR #131, opened 2019-10-11
Bn

bjrnfrdnnd:
Added the ability to import regex if that can be imported.
Regex is a drop-in replacement for re.
Improvement in regex: ability to define fuzzy regular expressions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,6 +6,9 @@ PyParsing -- A Python Parsing Module
 Introduction
 ============
 
+This version of pyparsing imports regex as re if regex is installed.
+It therefore allows for the definition of expressions that can be matched in a fuzzy way.
+
 The pyparsing module is an alternative approach to creating and
 executing simple grammars, vs. the traditional lex/yacc approach, or the
 use of regular expressions. The pyparsing module provides a library of
--- a/pyparsing.py
+++ b/pyparsing.py
@@ -117,8 +117,8 @@ if sys.version_info >= (3, 0) and sys.version_info <= (3,3):
     regex_loader = importlib.find_loader('regex')
     regex_found = regex_loader is not None
 if sys.version_info >= (3, 4):
-    import importlib
-    regex_spec = importlib.util.find_spec("regex")
+    from importlib.util import find_spec
+    regex_spec = find_spec("regex")
     regex_found = regex_spec is not None
 if regex_found:
     import regex as re
--- a/pyparsing.py
+++ b/pyparsing.py
@@ -104,7 +104,26 @@ from weakref import ref as wkref
 import copy
 import sys
 import warnings
-import re
+regex_found = False
+if sys.version_info >= (2, 0) and sys.version_info < (3,0):
+    import imp
+    try:
+        imp.find_module('eggs')
+        regex_found = True
+    except ImportError:
+        regex_found = False
+if sys.version_info >= (3, 0) and sys.version_info <= (3,3):
+    import importlib
+    regex_loader = importlib.find_loader('regex')
+    regex_found = regex_loader is not None
+if sys.version_info >= (3, 4):
+    import importlib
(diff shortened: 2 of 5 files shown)


pyparsing.py
```diff
@@ -104,7 +104,26 @@
...
+    regex_loader = importlib.find_loader('regex')
+    regex_found = regex_loader is not None
+if sys.version_info >= (3, 4):
+    from importlib.util import find_spec
+    regex_spec = find_spec("regex")
+    regex_found = regex_spec is not None
+if regex_found:
+    import regex as re
+else:
+    import re
```
klahnakoski: Can you reduce these lines to:

```python
try:
    import regex as re
except:
    import re
```

README.rst
```diff
@@ -6,6 +6,9 @@ PyParsing -- A Python Parsing Module
 Introduction
 ============
 
+This version of pyparsing imports regex as re if regex is installed.
+It therefore allows for the definition of expressions that can be matched in a fuzzy way.
+
```
klahnakoski: Remove this comment

bjrnfrdnnd:
Ah Good solution

That solves the Problem at the cost of one additional line upon Import. 
I suggest you add a comment regarding that workaround in the docs. 

Thanks Bjoern

pyparsing.py
ptmcg: Not fond of bare `except:`s.  We could do `except ImportError:`, or just `except Exception:` to guard against other exceptions that might crop up in the regex module itself.

bjrnfrdnnd:
﻿Oh fine with me. 
Should I change that and do a new pull request? 
Kyle is not fond of the whole idea as you can do
import regex 
Import pyparsing 
pyparsing.re = regex

I like it better my way as it is only one line instead of three upon import

Bye Björn

bjrnfrdnnd:
Ups finally I do not understand as I did
except ImportError:

klahnakoski:
@bjrnfrdnnd   When deciding which way to go, this is @ptmcg's project.  I am just some other guy interested in the project.

ptmcg:
I'm reluctant to have pyparsing use the `regex` module simply because it happens to be installed. People get things pipped into their environments for any number of reasons, often because they were a hidden dependency of something else. The only time I've done something like this in the past was to import the backport of OrderedDict for the packrat enhancements, and this was for a backport, not an alternative module.

Pyparsing's Regex class currently accepts strings _or_ compiled re's for its constructor. Would you find it acceptable for Regex to take a regex-compiled re? (I tested this, and currently the regex-compiled re fails - this is very easily remedied.)  Your code would look like:

    import pyparsing as pp
    import regex

    integer_parser = pp.Regex(regex.compile(r'\d+'))   # would use regex for this expression

You would only need to do this for patterns that really need `regex` processing.

I also tried the suggestion by @klahnakoski (needs a minor change in pyparsing, adding 're' to '__all__'):

    import pyparsing as pp
    import regex
    pp.re = regex  # redirects all internal re usage in pyparsing to regex

    integer_parser = pp.Regex(r'\d+')  # would now use regex instead of re to compile this string
    integer_parser = pp.Word(pp.nums)  # would also now use regex internally instead of re

Either of these works for me.

-- Paul

bjrnfrdnnd:
Fine with me. 
Both workarounds are reasonable. 
Thanks,Bjoern

ptmcg:
Bjorn - I have pushed these changes up to the master branch for the next release. It turns out that either of the approaches I suggested will work with these changes. Thanks for submitting this idea, I think it is a nice addition to pyparsing.

bjrnfrdnnd:
Great thanks

And of course a try except clause in 4 lines is possible. 
I basically did not want to try to import Regex; mostly because I wanted to try out the different tests for the different python versions.
But maybe there is no downside whatsoever to the 4 line solution 

Thanks,Bjoern

PR #132, opened 2019-10-11
First cut at refactor pyparsing across multiple source modules

ptmcg:
After briefly looking at this PR's change set, I may want to move `__diag__` to core.py, since it is most directly used in that code.


ptmcg:
Looking at the backport (or more like foreport) of the change for support of the 3rd party regex module, I can see how allow for constructing Regex instances with regex-compiled expressions. But unless I'm mistaken, the replacement of the imported `re` using `import pyparsing; pyparsing.re = regex` will now require `import pyparsing.core; pyparsing.core.re = regex`. I'm not keen on exposing the refactored modules below the pyparsing module. Am I missing something simple?

ptmcg:
I've decided to address the inability to do `pyparsing.re = regex` by removing the mention of that option from the CHANGES docs, and focus only on the ability to use `regex`-compiled objects when creating `Regex` expressions.

klahnakoski:
on the subject of `import pyparsing; pyparsing.re = regex` there are a few options:

1. Make a function that performs the replacment of `re`:  For example `pyparsing.use(Regex)` will ensure `re` is replaced with `Regex` everywhere
2. Make yet-another-option switch: `pyparsing.use_regex=True` (or put something in the `__compat__` object)
3. Ensuring `pyparsing.re` is used in place of plain `re` in all modules - then `pyparsing/__init__.py` does the `re` import, and `import pyparsing; pyparsing.re = regex` can be used to replace `re` 
3. Put the `try: import Regex as re; except: pass` ([or the long verison](https://github.com/pyparsing/pyparsing/pull/131/files#diff-ea9b784481a02f9eac5b2944fab65a9e)) in a `compatibility.py`module.

ptmcg:
@klahnakoski I came up with pretty much the same set of options, and I don't really like any of them. In the end I prefer leaving the `regex` injection and dependency to be in the caller code, not in pyparsing itself. Otherwise I think it opens the door to requests to support other alternative modules (arrow for datetime, itertoolz for itertools, etc.). I am also sensitive to how bloated pyparsing's API has gotten, so adding more methods like `use()` or `use_re_module()` just adds to this problem.

Plus the OP already agreed that the less-invasive version that I am using now (accepting regex-compiled expressions when creating a Regex) is sufficient for their needs, so adding support for the blanket override of `re` with `regex` is not an urgent matter that needs tackling now.

ptmcg:
I could use some help here, as I am mystified why TravisCI is failing the unit tests, when they work fine in my PyCharm environment. `testRepeater` seems to be the problem child. I would prefer to merge a working PR, instead of a broken one to be fixed up post-merge.

ptmcg:
Fixed - in my merges the updates to the context restorer got lost.

klahnakoski:
@ptmcg  

I added your fork as a remote in my dev setup:  I am going to tackle breaking up the `pyparsing/core.py` into more modules. I will make a PR on your fork if/when I am done, and you can push from there to pyparsing.

Since the merge did not go well, are you still moving forward with this branch?  Are you going to make another attempt so that git can better merge changes from other branches?  Maybe you do not need to decide now: Once the refactoring is "good enough"  we can attempt to cleanup the change history so git can translate the changes between branches well-enough.

ptmcg:
This PR is now mergeable, so go ahead and work off the corresponding fork/branch.

ptmcg:
Closing, will rework with new fork from latest

PR #133, opened 2019-10-16
Support building without setuptools

scarabeusiv:
As pyparsing is a dependency used by setuptools this allows packagers to use distutils as a fallback when bootstraping setuptools without much fuss.

The package will be built it will just render the setup.py test not working for anyone not building with setuptools, but that is just a trivial issue i would say.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,10 @@
 
 """Setup script for the pyparsing module distribution."""
 
-from setuptools import setup
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils.core import setup
 from pyparsing import __version__ as pyparsing_version, __doc__ as pyparsing_description
 
 modules = ["pyparsing",]


ptmcg:
This exact code used to be in setup.py, but was removed in 2.4.2 - see the notes with this change: https://github.com/pyparsing/pyparsing/commit/b0310f8d52f8cb25ce3990e8130e1ed0250f20c1#diff-2eeaed663bd0d25b7e608891384b7298

scarabeusiv:
Yeah but this way distributions have to patch it as we really are not keen on bundled stuff. I can keep the local patch in openSUSE, but as it is just 3 lines and does not harm anyone I thought nobody woudl be too much against that.

jdufresne:
> Yeah but this way distributions have to patch it as we really are not keen on bundled stuff. 

Does this mean the distro will patch setuptools to unbundle its dependencies, but not patch pyparsing? Why the distinction?

distutils does not support some of the keywords in use. For example, `python_requires`. Running this script with distutils instead of setuptools results in:

```
/usr/lib64/python3.7/distutils/dist.py:274: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
```

> but as it is just 3 lines ...

Ultimately, I agree so feel free to revert.

scarabeusiv:
Well the setuptools project would not merge huge chunk like that that changes something they do on purpose, but as here it is 3 liner it kinda makes sense to make it work, as it breaks nothing, it just allows you to 'accidentaly' build it on unsupported python as it does not recognise the limiter and you can't run 'python setup.py test'.

You can see here it builds happily without setuptools in sight :)
  https://build.opensuse.org/package/show/devel:languages:python/python-pyparsing

ptmcg:
I will err on the side of flexibility here.

PR #137, opened 2019-10-17
Remove unused setup.py keywords "test_suite"

jdufresne:
The tests are run through either tox or Travis CI. In both cases, the
setup.py "test_suite" entrypoint is not used.

Further, the use of "test_suite" is discouraged as it installs using
eggs rather than pip.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -33,6 +33,11 @@ Version 3.0.0a1
     to help identify those expressions in your parsers that
     will have changed as a result.
 
+- Removed support for running `python setup.py test`. The setuptools
+  maintainers consider the test command deprecated (see
+  <https://github.com/pypa/setuptools/issues/1684>). To run the Pyparsing test,
+  use the command `tox`.
+
 - POTENTIAL API CHANGE:
   ZeroOrMore expressions that have results names will now
   include empty lists for their name if no matches are found.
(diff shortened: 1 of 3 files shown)


jdufresne:
Why keep it? it is unused.

Locally, to test one can use `tox` or `tox -e py38`. On Travis, the `.travis.yml` file does its own thing. So I don't see the point of adding a different test entry point.

Are you using this now? It doesn't run the `simple_unit_tests.py` script.

klahnakoski:
@ptmcg My understanding is we are keeping this line, but I am also glad to be wrong.  I defer to your authority

ptmcg:
I had a request to support "python setup.py test" (though I can't track it down at the moment), and this was the change that was needed to support that.  This looks like a pretty standard way for people to run  tests on packages they've pulled from GitHub, in addition to other available test mechanisms like tox, et al.  I'm inclined to leave it, and I think Kyle's changes pull in some other tests too(?).

jdufresne:
The setuptools maintainers recommended against using the `python setup.py test` and the `test_suite` keyword. They consider them both deprecated. They recommend instead to use tox, which is what I'm suggesting with this change -- that we follow their recommendation.

Here are their statements on the matter:

https://github.com/pypa/setuptools/issues/931
https://github.com/pypa/setuptools/issues/1684

ptmcg:
Would be good to add a note to CHANGES file that 'python setup.py test' is no longer supported, with the pypa links or quotes included.

jdufresne:
Done. Added an entry to CHANGES. I also added `tox.ini` to the sdist through `MANIFEST.in` so distro packagers can continue to run tests with a single command.

ptmcg:
Please resolve the conflicts on this PR, then I can merge it.

jdufresne:
Done

PR #138, opened 2019-10-17
Use pyupgrade to upgrade the code to use Python3 conventions

jdufresne:
The pyupgrade project is available at
https://github.com/asottile/pyupgrade and can be installed through pip.

The pyupgrade tool automatically upgrades syntax for newer versions of
the language. As pyparsing is now Python 3 only, can apply some cleanups
and simplifications. Ran the tool using the following command:

    $ find . -name \*.py -exec pyupgrade --py3-plus {} \;

### DIFF
--- a/examples/invRegex.py
+++ b/examples/invRegex.py
@@ -17,7 +17,7 @@ from pyparsing import (Literal, oneOf, printables, ParserElement, Combine,
     SkipTo, infixNotation, ParseFatalException, Word, nums, opAssoc,
     Suppress, ParseResults, srange)
 
-class CharacterRangeEmitter(object):
+class CharacterRangeEmitter:
     def __init__(self,chars):
         # remove duplicate chars in character range, but preserve original order
         seen = set()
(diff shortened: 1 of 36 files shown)


examples/invRegex.py
```diff
@@ -28,56 +28,50 @@ def __repr__(self):
...
-class AlternativeEmitter(object):
+class AlternativeEmitter:
     def __init__(self,exprs):
         self.exprs = exprs
     def makeGenerator(self):
         def altGen():
             for e in self.exprs:
-                for s in e.makeGenerator()():
-                    yield s
+                yield from e.makeGenerator()()
```
klahnakoski: I always wanted this in py2.  If this change gets accepted, then I guess all my other arguments are mute.

ptmcg:
Could you please remove the pyparsing.py changes from this PR? It looks like all the rest will merge fine, but the pyparsing.py refactor is almost done, and then you can run pyupgrade on the refactored files.

jdufresne:
No problem. Done in the latest revision.

PR #143, opened 2019-10-17
Py3 cleanup: Remove workaround for Python2 urllib

jdufresne:
For Python 3 only code, the import path is known and stable. Can remove
the ImportError workaround.

### DIFF
--- a/examples/statemachine/statemachine.py
+++ b/examples/statemachine/statemachine.py
@@ -8,14 +8,7 @@ import sys
 import os
 import types
 import importlib
-try:
-    import urllib.parse
-    url_parse = urllib.parse.urlparse
-except ImportError:
-    print("import error, Python 2 not supported")
-    raise
-    import urllib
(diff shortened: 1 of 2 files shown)


examples/statemachine/statemachine.py
```diff
@@ -8,14 +8,7 @@
...
 import importlib
-try:
-    import urllib.parse
-    url_parse = urllib.parse.urlparse
-except ImportError:
-    print("import error, Python 2 not supported")
-    raise
-    import urllib
-    url_parse = urllib.parse
+from urllib.parse import urlparse
```
klahnakoski: importing from a py2/py3 module would be better.

I agree that the try/except is a mess

klahnakoski:
@jdufresne @ptmcg   Discuss here:   https://github.com/orgs/pyparsing/teams/pyparsing-dev/discussions/7

PR #144, opened 2019-10-17
Py3 cleanup: Remove workaround from Python3 unichr()

jdufresne:
On Python3, always use chr().

### DIFF
--- a/examples/booleansearchparser.py
+++ b/examples/booleansearchparser.py
@@ -87,11 +87,6 @@ from __future__ import print_function
 from pyparsing import Word, alphanums, Keyword, Group, Forward, Suppress, OneOrMore, oneOf
 import re
 
-# Py2 compatibility
-try:
-    _unichr = unichr
-except NameError:
-    _unichr = chr
 
 alphabet_ranges = [
     ##CYRILIC: https://en.wikipedia.org/wiki/Cyrillic_(Unicode_block)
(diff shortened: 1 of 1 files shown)


examples/booleansearchparser.py
```diff
@@ -151,7 +146,7 @@ def parser(self):
 
         #suport for non-western alphabets
         for r in alphabet_ranges:
-            alphabet += u''.join(_unichr(c) for c in range(*r) if not _unichr(c).isspace())
+            alphabet += u''.join(chr(c) for c in range(*r) if not chr(c).isspace())
```
klahnakoski: @ptmcg, @jdufresne   I agree with `chr`, but I believe it is still premature to remove all py2.  At least keep the imports at the top of the file...

```
    if PY2:
        chr = unichr
```

or maybe use `six` or some future (now past) compatibility lib?

jdufresne:
> I believe it is still premature to remove all py2

Per [`setup.py`](https://github.com/pyparsing/pyparsing/blob/master/setup.py#L21), this package requires Python ≥ 3.5. So that compatability shim will not accomplish anything.

It is also documented that Python 2 is not supported.

https://github.com/pyparsing/pyparsing/blob/719c9a50d9995f067311a3dee5e19c2354df1049/CHANGES#L7-L9

ptmcg:
I am less concerned about Py2-compatible bits remaining in some of the examples. That said, Py2 versions of the examples can always be retrieved from the 2.4.x branch.

jdufresne:
I think the examples should follow the same best practices as the code itself. If the code is Python-3-only, then I think it makes sense to provide the smoothest possible examples for users that show how to use the library in a Python 3 environment. Additional noise through compatibility shims don't help the user understand the meat of the example.

As you say, Python2 users can continue using examples from the Python2 release.

PR #145, opened 2019-10-17
Py3 cleanup: Remove use of closing() with urlopen()

jdufresne:
In Python 3, urlopen() can always be used as a context manager. Wrapping
with closing() is not necessary.

https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen

> This function always returns an object which can work as a context
> manager …

### DIFF
--- a/examples/htmlStripper.py
+++ b/examples/htmlStripper.py
@@ -6,8 +6,8 @@
 #
 # Copyright (c) 2006, 2016, Paul McGuire
 #
-from contextlib import closing
-import urllib.request, urllib.parse, urllib.error
+import urllib.parse, urllib.error
+from urllib.request import urlopen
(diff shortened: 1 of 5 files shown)


examples/htmlStripper.py
```diff
@@ -17,7 +16,7 @@
 
 # get some HTML
 targetURL = "https://wiki.python.org/moin/PythonDecoratorLibrary"
-with closing(urllib.request.urlopen( targetURL )) as targetPage:
+with urllib.request.urlopen( targetURL ) as targetPage:
```
klahnakoski: Can we have `urlopen(...)` everywhere instead of the verbose `urllib.request.urlopen(...)`

jdufresne: Yup! Can do.

PR #147, opened 2019-10-18
Remove unused imports

jdufresne:
Discovered using the command:

    flake8 --select F401 .

### DIFF
--- a/examples/SimpleCalc.py
+++ b/examples/SimpleCalc.py
@@ -33,7 +33,7 @@ debug_flag=False
 
 variables = {}
 
-from fourFn import BNF, exprStack, fn, opn, evaluate_stack
+from fourFn import BNF, exprStack, evaluate_stack
 
 # from fourFn import BNF, exprStack, fn, opn
 # def evaluateStack( s ):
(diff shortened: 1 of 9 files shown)


examples/statemachine/documentSignoffDemo.py
```diff
@@ -4,7 +4,6 @@
 # Example of a state machine modeling the state of a document in a document
 # control system, using named state transitions
 #
-import statemachine
```
paulmcg-iio: Even though there are no references to the statemachine module, it must be imported in order for the .pystate modules to be importable.

jdufresne:
@paulmcg-iio did you remove your previous comment? Does it no longer apply or should I go ahead and revert the `statemachine` imports?

ptmcg:
I removed the comment because I accidentally used my work GH account instead of my personal account. The body of the comment is still valid - all the statemachine demos need to `import statemachine`, which will define the import hooks to support importing `.pystate` modules.

jdufresne:
Thanks! I have restored the `statemachine` imports in the latest revision.

PR #157, opened 2019-10-28
select_parser example: misc improvements

rcoup:
Update for the select_parser example:

* sqlite now supports `TRUE` and `FALSE` as literal values
* use pyparsing_common for numeric expressions
* fix identifier quoting (`"`, with escaping via `""`)
* downcase identifiers unless they're quoted
* fix string quoting (escape with `''`)
* add support for sql comments (`-- comment`)
* additional test cases

One thing I would like to quite figure out is where failures for these test cases (not included here) are coming from, must be something to do with operator precedence but I can't spot where:

```
SELECT * FROM abcd WHERE ff NOT IN (1,2,4,5)
                            ^
FAIL: Expected end of text, found 'N'  (at char 28), (line:1, col:29)
SELECT * FROM abcd WHERE ff not between 3 and 9
                            ^
FAIL: Expected end of text, found 'n'  (at char 28), (line:1, col:29)
SELECT * FROM abcd WHERE ff not like 'bob%'
                            ^
FAIL: Expected end of text, found 'n'  (at char 28), (line:1, col:29)
```

### DIFF
--- a/examples/select_parser.py
+++ b/examples/select_parser.py
@@ -13,132 +13,24 @@ DOT, STAR = map(Literal, ".*")
 select_stmt = Forward().setName("select statement")
 
 # keywords
-(
-    UNION,
-    ALL,
-    AND,
-    INTERSECT,
-    EXCEPT,
-    COLLATE,
-    ASC,
-    DESC,
-    ON,
-    USING,
-    NATURAL,
-    INNER,
-    CROSS,
-    LEFT,
-    OUTER,
-    JOIN,
-    AS,
-    INDEXED,
-    NOT,
-    SELECT,
-    DISTINCT,
-    FROM,
-    WHERE,
-    GROUP,
-    BY,
-    HAVING,
-    ORDER,
-    BY,
-    LIMIT,
-    OFFSET,
-    OR,
-) = map(
-    CaselessKeyword,
-    """UNION, ALL, AND, INTERSECT,
- EXCEPT, COLLATE, ASC, DESC, ON, USING, NATURAL, INNER, CROSS, LEFT, OUTER, JOIN, AS, INDEXED, NOT, SELECT,
- DISTINCT, FROM, WHERE, GROUP, BY, HAVING, ORDER, BY, LIMIT, OFFSET, OR""".replace(
-        ",", ""
-    ).split(),
-)
-(
-    CAST,
-    ISNULL,
-    NOTNULL,
-    NULL,
-    IS,
-    BETWEEN,
-    ELSE,
-    END,
-    CASE,
-    WHEN,
-    THEN,
-    EXISTS,
-    IN,
-    LIKE,
-    GLOB,
-    REGEXP,
-    MATCH,
-    ESCAPE,
-    CURRENT_TIME,
-    CURRENT_DATE,
-    CURRENT_TIMESTAMP,
-) = map(
-    CaselessKeyword,
-    """CAST, ISNULL, NOTNULL, NULL, IS, BETWEEN, ELSE, END, CASE, WHEN, THEN, EXISTS, IN, LIKE, GLOB,
- REGEXP, MATCH, ESCAPE, CURRENT_TIME, CURRENT_DATE, CURRENT_TIMESTAMP""".replace(
-        ",", ""
-    ).split(),
-)
-keyword = MatchFirst(
-    (
-        UNION,
-        ALL,
-        INTERSECT,
-        EXCEPT,
-        COLLATE,
-        ASC,
-        DESC,
-        ON,
-        USING,
-        NATURAL,
-        INNER,
-        CROSS,
-        LEFT,
-        OUTER,
-        JOIN,
-        AS,
-        INDEXED,
-        NOT,
-        SELECT,
-        DISTINCT,
-        FROM,
-        WHERE,
-        GROUP,
-        BY,
-        HAVING,
-        ORDER,
-        BY,
-        LIMIT,
-        OFFSET,
-        CAST,
-        ISNULL,
-        NOTNULL,
-        NULL,
-        IS,
-        BETWEEN,
-        ELSE,
-        END,
-        CASE,
-        WHEN,
-        THEN,
-        EXISTS,
-        COLLATE,
-        IN,
-        LIKE,
-        GLOB,
-        REGEXP,
-        MATCH,
-        ESCAPE,
-        CURRENT_TIME,
-        CURRENT_DATE,
-        CURRENT_TIMESTAMP,
-    )
-)
+keywords = {
+    k: CaselessKeyword(k)
+    for k in """\
+    UNION ALL AND INTERSECT EXCEPT COLLATE ASC DESC ON USING NATURAL INNER CROSS LEFT OUTER JOIN AS INDEXED NOT
+    SELECT DISTINCT FROM WHERE GROUP BY HAVING ORDER LIMIT OFFSET OR CAST ISNULL NOTNULL NULL IS BETWEEN ELSE END
+    CASE WHEN THEN EXISTS IN LIKE GLOB REGEXP MATCH ESCAPE CURRENT_TIME CURRENT_DATE CURRENT_TIMESTAMP TRUE FALSE
+    """.split()
+}
+vars().update(keywords)
+
+keyword = MatchFirst(keywords.values())
 
 identifier = ~keyword + Word(alphas, alphanums + "_")
+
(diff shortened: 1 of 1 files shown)


examples/select_parser.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
ptmcg: This is the default encoding for Py3 now, we won't be adding back any Py2 compatibility code in core or examples.

ptmcg:
Your test failures come from the unexpected NOT as part of the IN, LIKE, and BETWEEN operators. This is  not the unary NOT, but a qualifier on one of the other operators. Try defining:

    OptNOT = Optional(NOT)

and then add `OptNOT +` as part of the IN, LIKE and BETWEEN (and any other NOT-able binary and ternary operators).

If this solves the problem, please add the `OptNOT` changes and the failing tests.

ptmcg:
I took a guess at which operators might be `NOT`-able. I don't think this is required for the unary `IN`, the unary `NOT` should work with that okay. If `GLOB` and `REGEXP` are also `NOT`-able, add top `OptNot` to them too:

    OptNOT = Optional(NOT)
    expr << infixNotation(expr_term,
        [
        (oneOf('- + ~') | NOT, UNARY, opAssoc.RIGHT),
        (ISNULL | NOTNULL | NOT + NULL, UNARY, opAssoc.LEFT),
        ('||', BINARY, opAssoc.LEFT),
        (oneOf('* / %'), BINARY, opAssoc.LEFT),
        (oneOf('+ -'), BINARY, opAssoc.LEFT),
        (oneOf('<< >> & |'), BINARY, opAssoc.LEFT),
        (oneOf('< <= > >='), BINARY, opAssoc.LEFT),
        (oneOf('= == != <>') | IS + OptNOT | OptNOT + IN | OptNOT + LIKE | GLOB | OptNOT + MATCH | REGEXP, BINARY, opAssoc.LEFT),
        ((OptNOT + BETWEEN,AND), TERNARY, opAssoc.LEFT),
        (IN + LPAR + Group(select_stmt | delimitedList(expr)) + RPAR, UNARY, opAssoc.LEFT),
        (AND, BINARY, opAssoc.LEFT),
        (OR, BINARY, opAssoc.LEFT),
        ])

ptmcg:
Oh, and thanks for submitting!

ptmcg:
I'm rethinking these "Optional(NOT)" changes. The problem with adding the "NOT" keywords in this way is that it will convert the single keyword operators into groups, even if no `NOT` is present. Parsing `LIKE` would now return the group `["LIKE"]`, `BETWEEN` as `["BETWEEN"]`, etc. 

So to retain the ability to detect "LIKE" as "LIKE", etc., these changes might be better:

    expr << infixNotation(expr_term,
        [
        (oneOf('- + ~') | NOT, UNARY, opAssoc.RIGHT),
        (ISNULL | NOTNULL | NOT + NULL, UNARY, opAssoc.LEFT),
        ('||', BINARY, opAssoc.LEFT),
        (oneOf('* / %'), BINARY, opAssoc.LEFT),
        (oneOf('+ -'), BINARY, opAssoc.LEFT),
        (oneOf('<< >> & |'), BINARY, opAssoc.LEFT),
        (oneOf('< <= > >='), BINARY, opAssoc.LEFT),
        (oneOf('= == != <>') | IS | IN | LIKE | GLOB | MATCH | REGEXP | NOT + (IN | LIKE | MATCH), BINARY, opAssoc.LEFT),
        ((BETWEEN | NOT + BETWEEN, AND), TERNARY, opAssoc.LEFT),
        (IN + LPAR + Group(select_stmt | delimitedList(expr)) + RPAR, UNARY, opAssoc.LEFT),
        (AND, BINARY, opAssoc.LEFT),
        (OR, BINARY, opAssoc.LEFT),
        ])

Now you will only get groups if there are multi-keyword operators, such as `["NOT", "LIKE"]`.

ptmcg:
I looked at this a little further, and there were several things going on. One was a typo in your `'bob%'` test, missing the trailing "'". Seems like this could be something added to QuotedString, to select whether a missing trailing quote should be treated as a fatal parse error - note to self for future release.

The other thing was that I didn't chase down all the places "NOT" could go. I got things pretty much working with this definition of expr:

```
NOT_NULL = Group(NOT + NULL)
NOT_BETWEEN = Group(NOT + BETWEEN)
NOT_IN = Group(NOT + IN)
NOT_LIKE = Group(NOT + LIKE)
NOT_MATCH = Group(NOT + MATCH)
NOT_GLOB = Group(NOT + GLOB)
NOT_REGEXP = Group(NOT + REGEXP)

UNARY,BINARY,TERNARY=1,2,3
expr << infixNotation(expr_term,
    [
    (oneOf('- + ~') | NOT, UNARY, opAssoc.RIGHT),
    (ISNULL | NOTNULL | NOT_NULL, UNARY, opAssoc.LEFT),
    ('||', BINARY, opAssoc.LEFT),
    (oneOf('* / %'), BINARY, opAssoc.LEFT),
    (oneOf('+ -'), BINARY, opAssoc.LEFT),
    (oneOf('<< >> & |'), BINARY, opAssoc.LEFT),
    (oneOf('< <= > >='), BINARY, opAssoc.LEFT),
    (oneOf('= == != <>') | IS | IN | LIKE | GLOB | MATCH | REGEXP 
                              | NOT_IN | NOT_LIKE | NOT_GLOB | NOT_MATCH | NOT_REGEXP, 
     BINARY, opAssoc.LEFT),
    ((BETWEEN | NOT_BETWEEN, AND), TERNARY, opAssoc.LEFT),
    ((IN | NOT_IN) + LPAR + Group(select_stmt | delimitedList(expr)) + RPAR, UNARY, opAssoc.LEFT),
    (AND, BINARY, opAssoc.LEFT),
    (OR, BINARY, opAssoc.LEFT),
    ])
```

This also takes care of all the operators being a single list or token, instead of sometimes being one word and sometimes being two (which makes iterating hard).

Lastly, I modified the end line to be better at summarizing the success/fail of the tests, and exiting with a 0 or 1 status code so this can be used as part of CI unit tests.

```
success, _ = select_stmt.runTests(tests)
print("\n{}".format("OK" if success else "FAIL"))
import sys
sys.exit(0 if success else 1)
```

examples/select_parser.py
ptmcg: (The emojis *are* a nice touch.)

rcoup:
Thanks @ptmcg, I'll fold these in :)

For me, at a conceptual level NOT is a unary operator, but `a NOT IN (1,2,3)` while _behaviourally equivalent in a database_ to `NOT (a IN (1,2,3))` is (as you propose) a better match as a binary NOT IN operator for the syntax/purposes of the parser.

Certainly when trying to do SQL → AST → SQL you produce the same thing you started with much more easily 😄

examples/select_parser.py
rcoup: IME by far the best way to test code for unicode-correctness 🦄

ptmcg:
Since running the Black PR, many files have changed format, so it may be easier to re-pull and reapply your changes (using Black formatting pre-commit).

Hmm, Black does not appear to like my style of defining large numbers of Keywords, I'll need to revisit this so as not to generate long, unflattering lists in the code.

ptmcg:
Here is a new statement for defining all the keywords that Black will leave alone. Has the added benefit of not requiring replicated keyword names and values. Downside is that PyCharm now complains about unknown symbols, since the keywords are assigned to the `vars()` dir dynamically:

```
keywords = {
    k: CaselessKeyword(k)
    for k in """\
    UNION ALL AND INTERSECT EXCEPT COLLATE ASC DESC ON USING NATURAL INNER CROSS LEFT OUTER JOIN AS INDEXED NOT 
    SELECT DISTINCT FROM WHERE GROUP BY HAVING ORDER LIMIT OFFSET OR CAST ISNULL NOTNULL NULL IS BETWEEN ELSE END
    CASE WHEN THEN EXISTS IN LIKE GLOB REGEXP MATCH ESCAPE CURRENT_TIME CURRENT_DATE CURRENT_TIMESTAMP
    """.split()
}
vars().update(keywords)

keyword = MatchFirst(keywords.values())

```

I'll probably start shifting over to this idiom moving forward, in both pyparsing and in my own personal development work.

rcoup:
@ptmcg sorry it took so long, those changes are all made and rebased/black'd onto master and CI is green. 

And the NOT test cases pass! 🙌 Thanks for your help with that, not sure I would have gotten there as elegantly ⭐️

PR #158, opened 2019-11-02
Use staticmethod decorator

jdufresne:
Available since Python 2.4.

https://docs.python.org/3/library/functions.html#staticmethod


ptmcg:
Wow, this code is really old. Kind of surprised Black didn't clear these up.

Merging...

PR #159, opened 2019-11-02
Use classmethod decorator

jdufresne:
Available since Python 2.4.

https://docs.python.org/3/library/functions.html#classmethod


ptmcg:
Dang! would have been nice to include with the staticmethod changes PR. Oh well.

Probably not necessary to link to Python docs for these features, especially since it is clear we only care about Py3 and beyond.

PR #163, opened 2019-11-06
Explicitly specify UTF-8 encoding when opening README file

twvd:
Hello,

On environments where the default encoding is set to ASCII (i.e. default Debian Stretch), installing pyparsing 2.4.4 fails with an error because README.rst contains a unicode character.
Explicitly specifying the encoding as UTF-8 fixes the problem.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@ Introduction
 ============
 
 The pyparsing module is an alternative approach to creating and
-executing simple grammars, vs. the traditional lex/yacc approach, or the
+executing simple grammars, vs. the traditional lex/yacc approach, or the
 use of regular expressions. The pyparsing module provides a library of
 classes that client code uses to construct the grammar directly in
 Python code.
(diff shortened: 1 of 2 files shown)


ptmcg:
Thanks for submitting this - pyparsing gets installed in so many different environments now. The upcoming 3.0 release will go through some more staged a, b, rc, etc. releases, so we should be able to work out these kind of issues before it gets released.

The `encoding` argument was not added to `open` until Py3, and this branch of pyparsing still supports Py2. We could wrap in a try-except for Py2-compat. Or the simpler solution is to just remove the offending character from README.rst - but I can't see what character that is.  Please update the PR to be Py2-compatible, either by try-except around the open, and/or removing the unicode character from README.rst.

twvd:
I found the character in README.rst (it was a space character) so I did both.
The ``from io import open`` should fix support for Python 2.6.x+ (see https://docs.python.org/2/library/io.html#io.open).

I'll leave it up to you if you want to merge all changes or just the stray character in README.rst :)

README.rst
```diff
@@ -7,7 +7,7 @@ Introduction
 ============
 
 The pyparsing module is an alternative approach to creating and
-executing simple grammars, vs. the traditional lex/yacc approach, or the
+executing simple grammars, vs. the traditional lex/yacc approach, or the
```
johnthagen: 👀

PR #173, opened 2020-01-15
Smedberg/various minor fixes

smedberg:
Various fixes to sample of parsing BigQuery SQL definition:

1. Failed parsing column specifications that include whitespace, like `tablename. columnname`
2. Failed parsing `UNION` that contains `WITH` statement
3. Previous `UNION` fix introduced a bug with parsing `SELECT`s that are surrounded with parenthesis
4. Table identifiers that are partially quoted and partially unquoted weren't properly handled, e.g. ``` `edw-prod-153420`.pdw.mrr_to_instance_mapping```


ptmcg:
Thanks for your continued work on this example.

PR #178, opened 2020-02-15
ensure test can fail

klahnakoski:
It seems this example was not failing if there was a problem

### DIFF
--- a/examples/eval_arith.py
+++ b/examples/eval_arith.py
@@ -163,107 +163,98 @@ comp_expr = infixNotation(
 )
 
 
-def main():
-    # sample expressions posted on comp.lang.python, asking for advice
-    # in safely evaluating them
-    rules = [
-        "( A - B ) = 0",
-        "(A + B + C + D + E + F + G + H + I) = J",
-        "(A + B + C + D + E + F + G + H) = I",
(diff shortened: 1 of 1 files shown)


klahnakoski:
View by ignoring whitespace (`?w=1`)

https://github.com/pyparsing/pyparsing/pull/178/files?w=1

ptmcg:
I was under the impression that returning something other than 0 constituted "fail", at least as far as TravisCI was concerned.

klahnakoski:
The problem is the interaction of the test suite with the example code.  I do not like this solution, because there are probaly other examples using the same pattern, so must be changed. Maybe you have a better suggestion.

Here is the old example code:

```
if __name__ == "__main__":
    exit(main())
```

The `test_examples.py` runs this example by importing the example:  

```
import_module("examples." + name)
```

which means the `__name__` is not `"__main__"`, and the example does not actually run.

By removing the `__name__` check, and flattening the example, the test suite can run it. Also, it will still run from the command line.  The only problem i see if if there are ParserElement artifacts you want to use in an example without running parse: I doubt that happens, but if it does then an example can be split into two files: One to build the parser, and a second to `import` that parser and run the example parse.

PR #179, opened 2020-02-15
Rename

klahnakoski:
This is the beginning of an attempt to streamline the ParseResults object.  I will add comments inline.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+  -   repo: https://github.com/python/black
+      rev: stable
+      hooks:
+        - id: black
+          language_version: python3.6
+          line_length: 88
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,0 @@
-repos:
-  -   repo: https://github.com/python/black
-      rev: stable
-      hooks:
-        - id: black
-          language_version: python3.6
-          line_length: 88
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -67,7 +67,7 @@ class ParseResults(object):
     """
 
     __slots__ = [
-        "_result__name", "_result__parent", "_result__all_names", "_result__as_list",
+        "_result__name", "_result__parent", "_result__all_names",
         "_result__modal", "_result__list", "_result__dict", "__weakref__"
     ]
 
@@ -78,7 +78,6 @@ class ParseResults(object):
         self._result__name = None
         self._result__parent = None
         self._result__all_names = {}
-        self._result__as_list = asList
         self._result__modal = modal
         if toklist is None:
             toklist = []
@@ -644,7 +643,7 @@ class ParseResults(object):
             self._result__parent = None
 
     def __getnewargs__(self):
-        return self._result__list, self._result__name, self._result__as_list, self._result__modal
+        return self._result__list, self._result__name, self._result__modal
 
     def __dir__(self):
         return dir(type(self)) + list(self.keys())
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -67,8 +67,13 @@ class ParseResults(object):
     """
 
     __slots__ = [
-        "_result__name", "_result__parent", "_result__all_names",
-        "_result__modal", "_result__list", "_result__dict", "__weakref__"
+        "_result__name",
+        "_result__parent",
+        "_result__all_names",
+        "_result__modal",
+        "_result__list",
+        "_result__dict",
+        "__weakref__",
     ]
 
     def __new__(cls, toklist=None, name=None, asList=True, modal=True):
@@ -635,7 +640,7 @@ class ParseResults(object):
     def __setstate__(self, state):
         self._result__list = state[0]
         self._result__dict, par, inAccumNames, self._result__name = state[1]
-        self._result__all_names =set(inAccumNames)
+        self._result__all_names = set(inAccumNames)
         if par is not None:
             self._result__parent = wkref(par)
         else:
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -71,8 +71,8 @@ class ParseResults(object):
         "_parent",
         "_all_names",
         "_modal",
-        "_list",
-        "_dict",
+        "_toklist",
+        "_tokdict",
         "__weakref__",
     ]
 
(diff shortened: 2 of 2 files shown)


pyparsing/results.py
```diff
@@ -66,41 +66,41 @@ def test(s, fn=repr):
         - year: 1999
     """
 
+    __slots__ = [
+        "_result__name", "_result__parent", "_result__all_names",
+        "_result__modal", "_result__list", "_result__dict", "__weakref__"
+    ]
+
```
klahnakoski: We will strictly declare the `__slots__` so that the objects are, hopefully, smaller and uniform.

pyparsing/results.py
```diff
@@ -66,41 +66,41 @@ def test(s, fn=repr):
...
+        if toklist is None:
+            toklist = []
+        if isinstance(toklist, list):
+            self._result__list = toklist[:]
+        elif isinstance(toklist, _generator_type):
+            self._result__list = list(toklist)
+        else:
+            self._result__list = [toklist]
+        self._result__dict = dict()
+        return self
```
klahnakoski: The attributes have been changed from their double underscore versions to one with a `_result__*` prefix.  This was done because double-underscore is meant for Python directives, so IDEs do not show them, which makes debugging pyparsing difficult, and also makes debugging `parseActions` in parsers difficult: It is difficult to browse the ParseResult objects.

The longer name has no impact because they are `__slots__` (aka indexes into a tuple)

pyparsing/results.py
```diff
@@ -66,41 +66,41 @@ def test(s, fn=repr):
...
-            if toklist is None:
-                toklist = []
-            if isinstance(toklist, list):
-                self.__toklist = toklist[:]
-            elif isinstance(toklist, _generator_type):
-                self.__toklist = list(toklist)
-            else:
-                self.__toklist = [toklist]
-            self.__tokdict = dict()
-
```
klahnakoski: Moved to `__new__` so we no longer need `__doInit`

pyparsing/results.py
```diff
@@ -66,41 +66,41 @@ def test(s, fn=repr):
...
         if name is not None and name:
             if not modal:
-                self.__accumNames[name] = 0
+                self._result__all_names = {name}
```
klahnakoski: Since `0` is the only value, we use a set

pyparsing/results.py
```diff
@@ -454,11 +454,11 @@ def copy(self):
...
         """
-        ret = ParseResults(self.__toklist)
-        ret.__tokdict.update(self.__tokdict)
-        ret.__parent = self.__parent
-        ret.__accumNames.update(self.__accumNames)
-        ret.__name = self.__name
+        ret = ParseResults(self._result__list)
+        ret._result__dict = dict(self._result__dict.items())
+        ret._result__parent = self._result__parent
+        ret._result__all_names |= self._result__all_names
```
klahnakoski: Notice we use set union operator

ptmcg:
I am completely in favor of removing all '__' names from the code, but the embedded '__' seems unnecessary. I think replacing with just leading single underscore is sufficient (i.e., from "__parent" to "_parent").

pyparsing/results.py
ptmcg: Using `__slots__` also has a memory benefit, which will be of great help in this class.

pyparsing/results.py
ptmcg: Great, been thinking about doing this for a while

pyparsing/results.py
ptmcg: Ha! That is how old this code is, pre-dates `set`. Then was kept to maintain compat with Python version before `set` (2.2? 2.3?). Nice catch!

klahnakoski:
>  I think replacing with just leading single underscore is sufficient (i.e., from "__parent" to "_parent").

I agree.  My use of `_result__` was an irrational fear of a namespace collision.  I will change it.

ptmcg:
I just now noticed that you asked for a review - I should have made my comments as part of that.

pyparsing/results.py
```diff
@@ -66,41 +66,46 @@ def test(s, fn=repr):
...
+            self._list = [toklist]
+        self._dict = dict()
+        return self
 
     # Performance tuning: we construct a *lot* of these, so keep this
     # constructor as small and fast as possible
     def __init__(
         self, toklist=None, name=None, asList=True, modal=True, isinstance=isinstance
     ):
-        if self.__doinit:
```
ptmcg: Please change `_result_xxx` names to just `_xxx`
Leave `_toklist` and `_tokdict`. Chopping to `_list` and `_dict` are not sufficiently descriptive (or expand to `_token_list` and `_token_name_dict` or `_results_name_dict` for more self-descriptive names)

pyparsing/results.py
```diff
@@ -66,41 +66,46 @@ def test(s, fn=repr):
         - year: 1999
     """
 
+    __slots__ = [
```
ptmcg: Use of `__slots__` is good

pyparsing/results.py
```diff
@@ -623,27 +624,26 @@ def pprint(self, *args, **kwargs):
...
+                self._name,
             ),
         )
 
     def __setstate__(self, state):
-        self.__toklist = state[0]
-        self.__tokdict, par, inAccumNames, self.__name = state[1]
-        self.__accumNames = {}
-        self.__accumNames.update(inAccumNames)
+        self._list = state[0]
```
ptmcg: Change this line and next to tuple unpack style.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
```
ptmcg: This overlaps changes in another PR

pyparsing/results.py
```diff
@@ -623,27 +628,25 @@ def pprint(self, *args, **kwargs):
...
         if par is not None:
-            self.__parent = wkref(par)
+            self._parent = wkref(par)
         else:
-            self.__parent = None
+            self._parent = None
 
     def __getnewargs__(self):
-        return self.__toklist, self.__name, self.__asList, self.__modal
+        return self._toklist, self._name, self._modal
```
ptmcg: Why is aslist arg removed here?

klahnakoski: Good catch.  I now removed this function.  It is used by pickle to create new ParseResults objects before their attributes are set.  Since no constructor code should be running when Pickle is recreating these objects, sending no parameters to `__new__` is the correct solution.

That said, this function is never run during testing.

.pre-commit-config.yaml
klahnakoski: my mistake.   fixed.

pyparsing/results.py
klahnakoski: I will remove this function in another PR

PR #180, opened 2020-02-15
Add Black formatting Git hook

klahnakoski:
This hook will run Black upon commit.  If something needs formatting, then the file is fixed, and the commit fails.  Commit a second time for success.  This only acts on changed files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,4 +4,4 @@ repos:
     hooks:
     - id: black
       language_version: python3.6
-      line_length: 90
+      line_length: 80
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.6
```
ptmcg: What are the implications of the language_version option here? I still think we are targetting 3.5 as our minimum Py3 minor version.

klahnakoski: This is the version of python used to run black; changing it tells the pre-commit to use another version of python to run black.

Black will not change the python compatibility of the file.

PR #181, opened 2020-02-15
Incomplete: Clean up packrat, remove debug code redundancy

klahnakoski:


klahnakoski:
This PR depends on another PR, so it will wait.  Comments inline

pyparsing/cache.py
```diff
@@ -0,0 +1,83 @@
...
+from threading import RLock
+
+
+@dataclass
+class PackratStats:
+    hit = 0
+    miss = 0
+
+
+class Cache:
```
klahnakoski: put the caching logic in `cache.py`

pyparsing/core.py
```diff
@@ -122,6 +121,13 @@ def enable_all_warnings(cls):
 # hide abstract class
 del __config_flags
 
+DebugActions = namedtuple("DebugActions", ["TRY", "MATCH", "FAIL"])
```
klahnakoski: Using a named tuple makes debugging a tiny bit easier: assigning names to the tuple indexes

pyparsing/core.py
```diff
@@ -300,7 +306,7 @@ def inlineLiteralsUsing(cls):
 
     def __init__(self, savelist=False):
         self.parseAction = list()
-        self.failAction = None
+        self.failAction = noop
```
klahnakoski: ensure `failAction` is always a function, then we need not do a type check (see below)

pyparsing/core.py
```diff
@@ -313,10 +319,14 @@ def __init__(self, savelist=False):
...
+        self.mayIndexError = (
+            True
+        )  # used to optimize exception handling for subclasses that don't advance parse index
         self.errmsg = ""
-        self.modalResults = True  # used to mark results names as modal (report only last) or cumulative (list all)
-        self.debugActions = (None, None, None)  # custom debug actions
+        self.modalResults = (
+            True
+        )  # used to mark results names as modal (report only last) or cumulative (list all)
+        self.debugActions = DebugActions(noop, noop, noop)  # custom debug actions
```
klahnakoski: `debugActions` are always a function, so type checks are not needed

pyparsing/core.py
```diff
@@ -558,80 +566,32 @@ def postParse(self, instring, loc, tokenlist):
 
     # ~ @profile
     def _parseNoCache(self, instring, loc, doActions=True, callPreParse=True):
-        TRY, MATCH, FAIL = 0, 1, 2
-        debugging = self.debug  # and doActions)
-
-        if debugging or self.failAction:
```
klahnakoski: Instead of having two near-identical sections, we will add this check *if* an exception happens

pyparsing/core.py
```diff
@@ -645,20 +605,21 @@ def _parseNoCache(self, instring, loc, doActions=True, callPreParse=True):
...
                             modal=self.modalResults,
                         )
-        if debugging:
-            # ~ print("Matched", self, "->", retTokens.asList())
-            if self.debugActions[MATCH]:
-                self.debugActions[MATCH](instring, tokensStart, loc, self, retTokens)
+            except Exception as err:
+                self.debug and self.debugActions.FAIL(instring, start, self, err)
+                raise
+        self.debug and self.debugActions.MATCH(instring, start, loc, self, retTokens)
```
klahnakoski: renamed `tokenStart` to `start` so this line is a bit shorter and Black keeps it on this line.

pyparsing/core.py
```diff
@@ -558,80 +566,32 @@ def postParse(self, instring, loc, tokenlist):
...
     # ~ @profile
     def _parseNoCache(self, instring, loc, doActions=True, callPreParse=True):
-        TRY, MATCH, FAIL = 0, 1, 2
-        debugging = self.debug  # and doActions)
-
-        if debugging or self.failAction:
-            # ~ print("Match", self, "at loc", loc, "(%d, %d)" % (lineno(loc, instring), col(loc, instring)))
-            if self.debugActions[TRY]:
-                self.debugActions[TRY](instring, loc, self)
+        self.debug and self.debugActions.TRY(instring, loc, self)
```
klahnakoski: You will see this pattern

    self.debug and self.debugActions.TRY(instring, loc, self)

which is the same as

     if self.debug:
        self.debugActions.TRY(instring, loc, self)

pyparsing/core.py
```diff
@@ -645,20 +605,21 @@ def _parseNoCache(self, instring, loc, doActions=True, callPreParse=True):
...
         return loc, retTokens
 
     def tryParse(self, instring, loc, raise_fatal=False):
-        try:
-            return self._parse(instring, loc, doActions=False)[0]
-        except ParseFatalException:
-            if raise_fatal:
-                raise
-            raise ParseException(instring, loc, self.errmsg, self)
+        with ParserElement.packrat_cache.lock:
```
klahnakoski: Instead of locking upon every `_parseCache` call, we lock it once when we enter into the parsing routine.

pyparsing/core.py
```diff
@@ -669,45 +630,34 @@ def canParseNext(self, instring, loc):
             return True
 
     # argument cache for optimizing repeated calls when backtracking through recursive expressions
-    packrat_cache = (
-        {}
-    )  # this is set later by enabledPackrat(); this is here so that resetCache() doesn't fail
-    packrat_cache_lock = RLock()
-    packrat_cache_stats = [0, 0]
+    packrat_cache = Cache()
```
klahnakoski: The `stats` and `lock` are now part of the `packrat_cache`

pyparsing/core.py
ptmcg: Would rather see as explicit:

    if self.debug:
        self.debugActions.FAIL(etc.)

pyparsing/core.py
ptmcg: Not a fan of this pattern.

pyparsing/core.py
ptmcg: What if packrat is not enabled? Would these unneeded lock/unlocks impact parse-time performance?

ptmcg:
Some of the ugly parts of pyparsing were done at the time in pursuit of parse-time performance, which is pyparsing's Achille's Heel. I don't care too much about performance at parser creation time, as this tends to be (but is not always) a one-time upfront task. So please keep this in mind when modifying code in any pre-parse, parse-impl or post-parse code, or in any ParseException or ParseResults construction code as these get created and destroyed a lot (cf. my attempts at pre-allocating Exceptions for each ParserElement).

ptmcg:
I was unkeen on the boolean-expression-in-place-of-explicit-if, but willing to consider it if it might yield a performance boost for parse-time code.

I did a quick test with dis and two methods, one using an explicit if-then statement, and one using "debug and function-call" boolean expression. Here is the test code:

    import dis

    def debug_with_if(debug, fn):
        if debug:
            fn()

    def debug_with_bool_expr(debug, fn):
        debug and fn()

    print("debug_with_if")
    dis.dis(debug_with_if)
    print()
    print("debug_with_bool_expr")
    dis.dis(debug_with_bool_expr)

Here is the output:

    debug_with_if
      4           0 LOAD_FAST                0 (debug)
                  2 POP_JUMP_IF_FALSE       10

      5           4 LOAD_FAST                1 (fn)
                  6 CALL_FUNCTION            0
                  8 POP_TOP
            >>   10 LOAD_CONST               0 (None)
                 12 RETURN_VALUE

    debug_with_bool_expr
      8           0 LOAD_FAST                0 (debug)
                  2 JUMP_IF_FALSE_OR_POP     8
                  4 LOAD_FAST                1 (fn)
                  6 CALL_FUNCTION            0
            >>    8 POP_TOP
                 10 LOAD_CONST               0 (None)
                 12 RETURN_VALUE

(Note: dis puts in the '>>' marker, I do not know its significance.)

They look very similar, but note that in the case of the if-statement code, the IF_FALSE jump (and in our code, the FALSE case will be much more common) goes on to the next statement, that of loading up None in preparation of returning (the implicit "return None" at the end of every function). In the bool-expression code, the IF_FALSE jump goes to the instruction before the next statement, that of popping the top value from some internal eval stack before moving on to the implied "return None". So using the boolean expression in place of "if debug:" will run one additional bytecode. This may not even be measureable, but even with optimization, at best it will just match the explicit if-then statement.

tl;dr - I prefer to stick with explicit "if debug:" statements over adopting "debug and ..." expressions.

klahnakoski:
Thank you!  I have very little experience with optimizing code for speed, so I appreciate your feedback, and I appreciate the time you spent going into detail.  You are the project owner; I am fine with you simply declaring "the right way".  I will push back if I disagree.

I will go back to using if-blocks.

> (Note: dis puts in the '>>' marker, I do not know its significance.)

I think it is showing the jump-point.

pyparsing/core.py
klahnakoski: Good point.  We need some stress tests to get better feedback about if changes improve performance or not.  The current test suite is not a good example because it is spending most of its time constructing grammars, not actually parsing.  If you know of some large examples, I can turn them into a stress test, with timing reporting, intended to be run manually.

The speed impact of this lock does depend on use-cases:  If there is a single large parse event, then this is small cost (and much smaller when packrat enabled).  If the user is calling `parseString` (or `tryParse`) many times on small input, then the cost of this lock is larger.  I do not know how often this latter scenario happens.

Even so, there was some discussion about parsing with multiple threads. This lock helps with that: allowing only one thing to be parsed at a time.

I am not familiar with all the entry points; I will need your guidance on that.  Maybe I will make 

```
@entrypoint
def parseString(...
```

That will perform the lock, and mark the entry points.

PR #185, opened 2020-02-22
INCOMPLETE: fix examples to work with unittest

klahnakoski:
This is what I am proposing so that all the examples can be run.  This depends on other PRs to be merged first


examples/LAparser.py
```diff
@@ -405,153 +405,21 @@ def parser(mo):
...
-    equation.runTests((t[1] for t in testcases), postParse=post_test)
-
-    ##TODO: Write testcases with invalid expressions and test that the expected
-    ## exceptions are raised.
-
-    print("Tests completed!")
-    print("PASSED" if all_passed[0] else "FAILED")
-    assert all_passed[0]
-
-
```
klahnakoski: Some examples are clearly split into parser-definition and test.  When this happens, an example is split into the two parts.

examples/SimpleCalc.py
```diff
@@ -32,7 +32,7 @@
 
 variables = {}
 
-from fourFn import BNF, exprStack, evaluate_stack
+from examples.fourFn import BNF, exprStack, evaluate_stack
```
klahnakoski: All examples must be run from the main pyparsing directory, not from the `examples` directory

examples/booleansearchparser.py
```diff
@@ -457,17 +456,11 @@ def Test(self):
 
             test_passed = sorted(matches) == sorted(_matches)
             if not test_passed:
-                print("Failed", repr(text), "expected", matches, "matched", _matches)
+                raise Exception(
+                    " ".join(
+                        ("Failed", repr(text), "expected", matches, "matched", _matches)
+                    )
+                )
```
klahnakoski: We will raise Exceptions instead of tracking "ok", or not

examples/pymicko.py
```diff
@@ -1609,27 +1609,11 @@ def program_end_action(self, text, loc, arg):
...
-            return self.rProgram.ignore(cStyleComment).parseFile(
-                filename, parseAll=True
-            )
-        except SemanticException as err:
-            print(err)
-            exit(3)
-        except ParseException as err:
-            print(err)
-            exit(3)
+        return self.rProgram.ignore(cStyleComment).parseFile(filename, parseAll=True)
```
klahnakoski: All the exception checks end up doing the same thing:  Removed.

examples/select_parser.py
```diff
@@ -188,49 +188,50 @@
...
+    SELECT * FROM abcd WHERE ff NOT IN (1,2,4,5)
+    SELECT * FROM abcd WHERE ff not between 3 and 9
+    SELECT * FROM abcd WHERE ff not like 'bob%'
+"""
+
+success, _ = select_stmt.runTests(tests)
+if success:
+    print("\n{}".format("OK"))
+else:
+    raise Exception("FAIL")
```
klahnakoski: Yes, I said I would throw errors instead of tracking success, but this way ends up with a smaller delta, and easier for you to review.

tests/test_examples.py
```diff
@@ -1,34 +1,58 @@
...
+#         self._run("oc")
+#
+#     def test_delta_time(self):
+#         self._run("delta_time")
+#
+#     def test_eval_arith(self):
+#         self._run("eval_arith")
+#
+#     def test_select_parser(self):
+#         self._run("select_parser")
```
klahnakoski: Should be removed

pyparsing/testing.py
```diff
@@ -3,192 +3,176 @@
...
-    __compat__,
-)
+from pyparsing.core import ParserElement, ParseException, Keyword, __diag__, __compat__
 
 
-class pyparsing_test:
+class reset_pyparsing_context:
     """
-    namespace class for classes useful in writing unit tests
-    """
```
klahnakoski: Python uses modules for namespaces.

tests/test_simple_unit.py
```diff
@@ -12,7 +12,9 @@
 from collections import namedtuple
 from datetime import datetime
 
-ppt = pp.pyparsing_test
+from pyparsing import testing
+
+ppt = testing
```
klahnakoski: we use the `pyparsing.testing` module instead of the `pyparsing.testing.pyparsing_test` class for namesapce.

ptmcg:
Are these still examples? Or have they become tests? My intention in adding examples to the test suite was less about getting more test coverage and more about ensuring that examples would still work after making changes in the core or changes in the API - an example that doesn't keep up with the code it is demonstrating isn't a very good example.

Generally, I'd like an example to be something easily run by itself, so that someone can "try it and see." This feels different when they have to run it by invoking pytest.

You've done a lot of work on this in this PR, but please hold off on further work until I understand how this changes the nature of this code.

klahnakoski:
> Generally, I'd like an example to be something easily run by itself, so that someone can "try it and see."

I agree, and that is still the case.  The changes still allow direct run from the command line, without the need for pytest.  Mostly, this PR is about removing the `__name__ == "__main__"` check so that pytest can run it as a test AND the removal of any `exit()` calls so pytest continues to run other tests after a failure.

PR #192, opened 2020-03-20
New unit tests

mattcarmody:
I've written some additional unit tests and came across a potential issue with matchPreviousExpr. Can you please take a look at it and let me know if this is expected behavior?


ptmcg:
I think the behavior is not an issue with matchPreviousExpr but with setResultsName. If so, it could be a major breaking change in 3.0.

After looking at your tests, I have some rewrites that I'll insert after merging your PR.

PR #193, opened 2020-03-22
Add OnlyOnce docstring, error message and unit test

mattcarmody:


mattcarmody:
Would you rather the error message be moved to a class attribute and called with self.errmsg? That seems to be a common theme now that I look at other classes after the fact.

ptmcg:
I'm not going to worry too much about consistency with other classes in core.py - I don't think this class even belongs here, should probably be in actions.py.

mattcarmody:
Ok sounds good. Would you like me to move it to actions.py?

ptmcg:
Yes, please make that change.

mattcarmody:
Maybe this is why it was in core.py still, but I'm having trouble moving it to actions because it calls _trim_arity from core.py. Importing from core.py in actions.py causes function calls in other places to be suddenly unrecognized. I'm not an expert on python imports, but it seems like it creates a circular reference and that is a problem. Any thoughts?

ptmcg:
Leave it where it is then, thanks for trying.

PR #197, opened 2020-03-29
Docstring formatting changes

mattcarmody:
When I was first using pyparsing I spent a lot of time reading the documentation and noted some formatting inconsistencies.

The first commit fixes list & code blocks that aren't formatted on read the docs the way they're intended to.
The second adds missing class references and inline code formatting.
The third tries to fix a broken reference but I wasn't able to test it to see if it works. It also demos a potential formatting change that is entirely your decision. Because read the docs has such a narrow page, a lot of inline code blocks require horizontal scrolling, sometimes only for comments which can be frustrating as a reader. If you prefer the current format I'll revert those, no problem!


ptmcg:
Looks good, thanks

PR #198, opened 2020-03-30
Add ParseResults unit tests

mattcarmody:


tests/test_unit.py
```diff
@@ -2366,6 +2366,182 @@ def testParseResultsWithNamedTuple(self):
...
+        print(result.dump())
+        self.assertParseResultsEquals(
+            result, expected, msg="issue with ParseResults.append()"
+        )
+
+    def testParseResultsClear(self):
+        """test simple case of ParseResults.clear()"""
+
+        tst = "spam eggs"
+        expr = pp.OneOrMore(pp.Word(pp.alphas))
```
ptmcg: Use same expr that you used in testParseResultsValues, so that you confirm that results names are cleared as well as the list.

tests/test_unit.py
```diff
@@ -2366,6 +2366,182 @@ def testParseResultsWithNamedTuple(self):
...
+    def testParseResultsValues(self):
+        """test simple case of ParseResults.values()"""
+
+        expr = pp.Word(pp.alphas)("first") + pp.Word(pp.alphas)("second")
+        result = expr.parseString("spam eggs")
+
+        values_list = [ii for ii in result.values()]
+        print(values_list)
+        msg = "issue calling ParseResults.values()"
+        self.assertIn("spam", values_list, msg=msg)
```
ptmcg: Better to test for `set(result.values()) == {"spam", "eggs"}`

tests/test_unit.py
```diff
@@ -2366,6 +2366,182 @@ def testParseResultsWithNamedTuple(self):
...
+            result, expected_dict={}, msg="issue with ParseResults.clear() as dict"
+        )
+
+    def testParseResultsExtendWithString(self):
+        """test ParseResults.extend() with input of type str"""
+
+        # use a parse action to append the reverse of the matched strings to make a palindrome
+        def make_palindrome(tokens):
+            tokens.extend(reversed([t[::-1] for t in tokens]))
+            return "".join(tokens)
```
ptmcg: I like the idea of this parse action, but please leave off the return statement, so that the assert asserts the expected value of the ParseResults directly, rather than infer correctness from a value computed from the ParseResults.

tests/test_unit.py
```diff
@@ -2366,6 +2366,182 @@ def testParseResultsWithNamedTuple(self):
...
+
+        print(result.dump())
+        self.assertParseResultsEquals(
+            result, ["spam", "eggs"], msg="issue with ParseResults before clear()"
+        )
+
+        result.clear()
+
+        print(result.dump())
+        self.assertParseResultsEquals(
```
ptmcg: You could merge these two calls to assertParseResultsEquals, and pass both expected_list and expected_dict args.

tests/test_unit.py
```diff
@@ -2366,6 +2366,182 @@ def testParseResultsWithNamedTuple(self):
...
+            name, dir_result, msg="name value wasn't returned by dir(ParseResults)"
+        )
+        self.assertIn(
+            "asList", dir_result, msg="asList was not returned by dir(ParseResults)"
+        )
+
+    def testParseResultsInsert(self):
+        """test ParseResults.insert() with named tokens"""
+
+        # use a parse action to insert the parse location in the front of the parsed results
```
ptmcg: This test seems a little light, only does one insert at location 0. Let's create a ParseResults by parsing "A B C D E F G H I J" into separate tokens, and create a list with the same contents. Then do 10 random inserts of "a b c d e f g h i j".split() at locations between -5 and 5 to both the ParseResults and the list. Then print them both and compare if they are equal.

    import pyparsing as pp
    from random import randint

    result = pp.Word(pp.alphas)[...].parseString("A B C D E F G H I J")
    compare_list = result.asList()

    print(result)
    print(compare_list)

    for s in "abcdefghij":
        index = randint(-5, 5)
        result.insert(index, s)
        compare_list.insert(index, s)

    print(result)
    print(compare_list)

mattcarmody:
Thanks for the feedback, the changes cleaned things up nicely. And made for a much more thorough test, thanks for writing that. I shamelessly stole those parse actions from the docstrings so I'm not surprised you like the idea!

PR #199, opened 2020-04-01
Parser Element and Keyword tests

mattcarmody:
These tests focus on previously uncovered paths. Two behaviors surprised me, though these are edge cases:

1. When a type other than ParserElement or string is used in any of the expressions (I use int as a test case) a warning is issued and the entire expr is set to None. This will later throw a ParseError when the expr is parsed because a None type can't be parsed.
2. When using type string with the & operator, it seems that can be parsed successfully provided they are in the order of the string being parsed, but not if they're out of order.

Both of these may be expected behavior, just thought I'd share.

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -2444,6 +2444,119 @@ class Test2_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
         ):
             result = expr.parseString("spam eggs")
 
+    def testParserElementMulOperatorWithTuples(self):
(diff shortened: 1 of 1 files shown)


mattcarmody:
Any feedback on these tests? If they're not thorough enough I can expand them.

ptmcg:
I don't know why I was holding off on this PR, so sorry for the delay. These are in fact exactly what I had in mind for adding coverage of these cases, thanks.

PR #209, opened 2020-04-30
Pop counter token and return rest instead of dropping all tokens in countedArray

yawor:
A really small change but a quite useful one.
The `countedArray` helper is very nice but it assumes that the count expression is directly preceding the list. I've stumbled upon a situation where there are some extra tokens in between the count token and items in the list.

Example: There are two flags right after the count value which describe the list in some way. They can take a single ascii letter or "-".
With my change, one can do:

```python
my_array = countedArray(Word(alphanums), And([
    Word(nums).setParseAction(lambda t: int(t[0])),
    Group(Char(alphas + "-") * 2)
]))

print(my_array.parseString("5 a b item1 item2 item3 item4 item5"))
```
which will print:
```python
[['a', 'b'], ['item1', 'item2', 'item3', 'item4', 'item5']]
```

This doesn't change a normal behaviour of `countedArray` where there's no `intExpr` provided or it returns only a single token.

### DIFF
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -55,7 +55,7 @@ def countedArray(expr, intExpr=None):
 
     def countFieldParseAction(s, l, t):
         n = t[0]
-        arrayExpr << And([expr] * n) if n else empty
+        arrayExpr << (And([expr] * n) if n else empty)
         # clear list contents, but keep any named results
         del t[:]
 
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -54,9 +54,9 @@ def countedArray(expr, intExpr=None):
     arrayExpr = Forward()
 
     def countFieldParseAction(s, l, t):
-        n = t[0]
+        n = t.pop(0)
         arrayExpr << (n and Group(And([expr] * n)) or Group(empty))
-        return []
+        return t
 
     if intExpr is None:
         intExpr = Word(nums).setParseAction(lambda t: int(t[0]))
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -55,7 +55,7 @@ def countedArray(expr, intExpr=None):
 
     def countFieldParseAction(s, l, t):
         n = t[0]
-        arrayExpr << (n and Group(And([expr] * n)) or Group(empty))
+        arrayExpr << And([expr] * n) if n else empty
         # clear list contents, but keep any named results
         del t[:]
 
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -54,9 +54,10 @@ def countedArray(expr, intExpr=None):
     arrayExpr = Forward()
 
     def countFieldParseAction(s, l, t):
-        n = t.pop(0)
+        n = t[0]
         arrayExpr << (n and Group(And([expr] * n)) or Group(empty))
-        return t
+        # clear list contents, but keep any named results
+        del t[:]
 
     if intExpr is None:
         intExpr = Word(nums).setParseAction(lambda t: int(t[0]))
--- a/tests/test_simple_unit.py
+++ b/tests/test_simple_unit.py
@@ -503,7 +503,7 @@ class TestCommonHelperExpressions(PyparsingExpressionTestCase):
         ),
         PpTestSpec(
             desc="A counted array of words",
-            expr=pp.countedArray(pp.Word("ab"))[...],
+            expr=pp.Group(pp.countedArray(pp.Word("ab")))[...],
             text="2 aaa bbb 0 3 abab bbaa abbab",
             expected_list=[["aaa", "bbb"], [], ["abab", "bbaa", "abbab"]],
         ),
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -2963,14 +2963,14 @@ class Test2_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
             print("got maximum excursion limit exception")
 
     def testCountedArray(self):
-        from pyparsing import Word, nums, OneOrMore, countedArray
+        from pyparsing import Word, nums, OneOrMore, Group, countedArray
 
         testString = "2 5 7 6 0 1 2 3 4 5 0 3 5 4 3"
 
         integer = Word(nums).setParseAction(lambda t: int(t[0]))
         countedField = countedArray(integer)
 
-        r = OneOrMore(countedField).parseString(testString)
+        r = OneOrMore(Group(countedField)).parseString(testString)
         print(testString)
         print(r)
 
@@ -2980,7 +2980,7 @@ class Test2_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
 
     # addresses bug raised by Ralf Vosseler
     def testCountedArrayTest2(self):
-        from pyparsing import Word, nums, OneOrMore, countedArray
+        from pyparsing import Word, nums, OneOrMore, Group, countedArray
 
         testString = "2 5 7 6 0 1 2 3 4 5 0 3 5 4 3"
 
@@ -2988,7 +2988,7 @@ class Test2_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
         countedField = countedArray(integer)
 
         dummy = Word("A")
-        r = OneOrMore(dummy ^ countedField).parseString(testString)
+        r = OneOrMore(Group(dummy ^ countedField)).parseString(testString)
         print(testString)
         print(r)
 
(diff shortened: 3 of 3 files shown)


ptmcg:
I am reluctant to change the expectation that this will always return just the array of items.

How about, if you *do* have some kind of structured data besides the count, then to access them after parsing you must give them results names, and the results names will be preserved on the returned list:

        import pyparsing as pp
        ppc = pp.pyparsing_common

        # array counter contains several fields
        # - number of elements
        # - type of elements
        # - source of elements
        counter_with_metadata = ppc.integer("count") + ppc.identifier("type") + ppc.identifier("source")

        countedField = pp.countedArray(pp.Word(pp.alphanums), intExpr=counter_with_metadata)

        testString = "5 string input item1 item2 item3 item4 item5"
        r = countedField.parseString(testString)

        print(testString)
        print(r.dump())
        print('type = {!r}'.format(r.type))
        print('source = {!r}'.format(r.source))

Prints:

```
5 string input item1 item2 item3 item4 item5
[['item1', 'item2', 'item3', 'item4', 'item5']]
- count: 5
- source: 'input'
- type: 'string'
type = 'string'
source = 'input'

```

yawor:
@ptmcg is this something that should work already or just an alternative solution which would need to be implemented?
I've tried this using current version (without my modification) and this doesn't seem to work right now, so I assume it's the latter. Also I think that your example wouldn't work as expected because the `identifier` helper uses `setName`, not `setResultsName`.
Anyway I agree that your approach is actually better.

ptmcg:
Yes, apologies - this is a proposed enhancement, and is not part of the currently released code. I wanted to confirm with you first that this would meet your requirements before going through the rest of the code check-in process, update CHANGES file, etc.

If you want to test it for yourself, replace the internal `countFieldParseAction` function with this:

```
    def countFieldParseAction(s, l, t):
        n = t.pop(0)
        if t.haskeys():
            copy_names = lambda ss, ll, tt, save_names=t._tokdict: tt._tokdict.update(save_names)
            arrayExpr << (n and Group(expr * n).addParseAction(copy_names) or Group(empty).addParseAction(copy_names))
        else:
            arrayExpr << (n and Group(expr * n) or Group(empty))
        return []

```

The use of `setName` vs `setResultsName` is a common question, especially since I rarely use the full `setResultsName` form any more. `setName` describes the expression itself, and gives a nice string in exceptions as opposed to the repr-like stuff that is the default.  `setResultsName` assigns the matched text found in the input string to the parse results as a named result. I added `setResultsName` *very* early in pyparsing's life, but the length of the name always bothered me. So I added a short-hand interface using Python's call syntax, so ` ppc.identifier("type")` is equivalent to ` ppc.identifier.setResultsName("type")`.

The code I pasted in the previous post is the guts of the corresponding unit test I'll add to go along with this feature. Very nice request by the way, it has come up once or twice before and I never really had a good solution, so this gave me another chance at putting a simple enhancement with backward compatibility and (hopefully) logical consistency.

I'll include this in the next 3.0.0a2 release.

yawor:
Actually I've started implementing your proposal myself. Here's my version of the implementation:
```python
    def countFieldParseAction(s, l, t):
        n = t[0]
        arrayExpr << (n and Group(And([expr] * n)) or Group(empty))

        if t.haskeys():
            r = ParseResults(asList=False)
            for k, v in t.items():
                r[k] = v
            return r
        return []
```
I don't know if this is better or worse. I've tested this and it seems to be working properly.

Regarding the identifier, now I see it uses `__call__` in `ParserElement`, which basically calls `setResultsName`. I didn't know this was implemented, that's why I haven't noticed this in your example. Nice shortcut, good to know :).

Also I've been thinking: would it be acceptable to add a `arrayResultsName=None` argument to `countedArray`, so it could also be potentially accessible by name instead of by index? It would be applied to arrayExpr:
```python
arrayExpr << (n and Group(And([expr] * n)) or Group(empty))(arrayResultsName)
```

ptmcg:
I'm glad you kept at this, the solution is actually even simpler.

In a parse action, you can modify the tokens that are passed in, and if you return None, then pyparsing will use the tokens that were passed in (as modified by the parse action). So if there are results names in the incoming `t` variable, there is no need to copy them to a return value, we can just clear out the list contents of `t` and leave any results names in place:

```
    def countFieldParseAction(s, l, t):
        n = t[0]
        arrayExpr << Group((expr * n) if n else Empty())
        # clear list contents, but keep any named results
        del t[:]
```
(I got rid of the call to pop() also, since the trailing del statement will take care of it.)

No need to add a results name, the caller can do that on the created expression.  There *is* still an extra layer of list nesting because of the Group, and if we take that out, things look cleaner. BUT it does add another API compatibility issue, since previous versions had the Group, and so post parsing code will have the [0] index that they will have to remove when they upgrade.

My bias has always been on backward compatibility, but for pyparsing, if not now, when?

(Once we settle this out, my intention is for you to take our collaborative code and insert to your PR, so you will get credit as a project contributor.)

yawor:
The solution with del looks great :).

Regarding the arrayResultsName addition, it doesn't change the indexed result. To access the list using indexes it still requires to use `result[0]`, so it doesn't seem to break backward compatibility (unless I've overlooked something). But in my tests if I apply the results name to arrayExpr directly I can then also access it by `result.somename`. This makes the code a little cleaner in parse actions for example. If I apply the results name to final expression returned by `countedArray`, then there's an extra layer of `And`, so to access the list I need to do `results.somename[0]`. Anyway this is not critical in any way and I won't commit this part if you still don't like it.

yawor:
I've found there are two ways of naming results in countedArray right now:
```python
my_array1 = pp.countedArray(pp.Word(pp.alphanums))("list")
r1 = my_array1.parseString("5 item1 item2 item3 item4 item5")
print(r1.dump())
print(r1.list[0])
```

```python
my_array2 = pp.countedArray(pp.Word(pp.alphanums).setResultsName("list", True))
r2 = my_array2.parseString("5 item1 item2 item3 item4 item5")
print(r2.dump())
print(r2[0].list)
```

In both cases the index is still required to access the list. If I combine both, I could access the list by
```python
r.list[0].list
```

If you don't want to add `arrayResultsName` then I think the first way is the best.

yawor:
@ptmcg I know this is outside of the scope of this PR, but I don't want to create a separate issue. I see that the current min version of Python is 3.5, so typing module and type hinting annotations are available (except variable type annotations, these were added in 3.6). I would like to add typing to pyparsing (it's very helpful in modern IDEs like PyCharm), but it's a lot of work, so before starting working on this I'd like to know if you'd be willing to accept such PR :).

ptmcg:
I think I'm going to break the backward compatibility here and remove the Group. This make things more consistent with delimitedList, which also does not Group, leaving the choice up to the caller to do so.  Its also looking like I'll need an API changes section at the top of the CHANGES file, since this 3.0.0 release involves a number of them.

I'm not keen on adding a parameter for defining a results name - I can't think of any other pyparsing ParserElement class or helper method that defines results names this way. And by removing the Group, it shouldn't be needed. Once the Group is removed, then results naming can be done in the calling code.  I'll need to test this out, but right now, I'm liking this approach the best.

Please update your PR with the del t[:] code, removing the Group. Don't bother writing a CHANGES blurb, I'll take care of that.

(As for type annotations, I am generally in favor of them. Are you sure there isn't an issue for this already? Maybe it was just on a todo list I was keeping for myself. They do get clunky when the code is lenient in what types are allowed, and gets messy with `Union[]`s, and pyparsing does a lot of duck-type style type inferences. And don't get me started on parse actions!)

pyparsing/helpers.py
```diff
@@ -55,7 +55,7 @@ def countedArray(expr, intExpr=None):
 
     def countFieldParseAction(s, l, t):
         n = t[0]
-        arrayExpr << And([expr] * n) if n else empty
+        arrayExpr << (And([expr] * n) if n else empty)
```
ptmcg, line 58: Can be written as just `arrayExpr << (expr * n) if n else empty` (the parens might not be needed, but I think they help readability).  I generally dislike using the explicit And constructions if they can be avoided., I much prefer using '+', '-', and '*' operators.

Otherwise looks good, good catch in fixing the unit tests! (another flag that we are changing the API...)

yawor: Actually, the expr * n was also my first choice, but I though that it was like that for some reason. Actually the `__mul__` in `ParserElement` has some overhead, so probably explicit `And` in this case may be faster (especially here, as this is called for each parsed "instance" of such array in parsed data), so I've left this unchanged.
Also the parentheses around are needed as left shift operator has lower precedence than ternary if/else.

ptmcg:
I still don't know when GitHub will ask for a comment or not. Here are the notes I was going to add on the review:

It just predated the '*' operator, that is all. Possibly a slight performance improvement to just use the And syntax, and this *does* happen at parse time, which is always worth optimizing. Let's keep as-is then.

Squash/merge forthcoming this afternoon, thanks for the effort!

PR #214, opened 2020-05-17
Add misc unit tests for core elements with missing coverage

mattcarmody:
I may be missing something, but it seems like the `__req__` and `__rne__` methods defined in core.py aren't built in dundees to override and aren't being used anywhere? I didn't have any success testing them with operators.


ptmcg:
Go ahead and delete them.

mattcarmody:
Done!

PR #215, opened 2020-05-17
Imports cleanup in unit tests

mattcarmody:
There are a bunch of imports in the unit tests that import pyparsing elements, some of which are exploded onto many lines by black. It seems unnecessary, since they can all be referenced with pp, ie `pp.Word`, and you encouraged me to use that format when writing new tests. I didn't change all the instances, but these tests caught my eye and felt cleaner without the imports.

This PR is mostly a matter of style and personal taste though, so feel free to disregard it.


ptmcg:
Looks good to me, those black-exploded lists and tuples are an eyesore.

PR #216, opened 2020-05-22
examples -- romanNumerals -- makeRomanNumeral bug fix, added MMMMM test

jayped007:
There was a bug in the makeRomanNumerals function in romanNumerals.py.  For example, makeRomanNumeral(2) would return 'I', and makeRomanNumeral(3) would return 'II', etc.  This was tracked to an issue within that function that was fixed.  Adding the test for 'MMMMM' caused a test failure with the previous code and now it works.  The 'MMMMM' test was added to the set of tests.


ptmcg:
Amazing! This bug has been there for about 14 years, nice catch!

PR #217, opened 2020-05-24
Add GoToColumn test

mattcarmody:
This class was previously untested so I adapted an example I found on Stack Overflow. It may be worth adding an example to the docstring.

I didn't have any luck exercising the two lines that still aren't covered in a useful way. I was trying to use `GoToColumn().ignore()` but it didn't behave the way I expected it to.

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -7542,6 +7542,47 @@ class Test2_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
             result, expected, msg="issue with OnlyOnce after reset"
         )
 
+    def testGoToColumn(self):
+        """tests for GoToColumn class"""
+
+        dateExpr = pp.Word(pp.nums + ".")("date")
+        realNum = pp.Word(pp.nums + ".")("real")
+        intNum = pp.Word(pp.nums + ".")("integer")
+        valueExpr = realNum | intNum
+        extraExpr = pp.OneOrMore(pp.Word(pp.alphanums))
+
+        sample = """\
+            date                Not Important                         value    NotImportant2
+            11.11.13            useless . useless,21 useless 2        14.21    asmdakldm
+            21.12.12            fmpaosmfpoamsp 4                      41       ajfa9si90""".splitlines()
+
+        # Column number finds match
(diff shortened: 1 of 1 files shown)


tests/test_unit.py
```diff
@@ -7542,6 +7542,47 @@ def append_sum(tokens):
             result, expected, msg="issue with OnlyOnce after reset"
         )
 
+    def testGoToColumn(self):
+        """tests for GoToColumn class"""
+
+        dateExpr = pp.Word(pp.nums + ".")("date")
+        realNum = pp.Word(pp.nums + ".")("real")
```
ptmcg, line 7549: Please use `real` and `integer` from `pyparsing_common` (or just use `pyparsing_common.number`), and change `dateExpr` to `Regex(r"\d\d(\.\d\d){2}")`. Defining reals as `Word(nums + '.')` is an anti-pattern that is listed in https://github.com/pyparsing/pyparsing/wiki/The-Classic-Blunders. It will match ".", "....", and many other strings that are not valid reals. I don't want to put out examples that will mislead those who might look through any of the test or example code for samples.

tests/test_unit.py
```diff
@@ -7542,6 +7542,47 @@ def append_sum(tokens):
...
+        intNum = pp.Word(pp.nums + ".")("integer")
+        valueExpr = realNum | intNum
+        extraExpr = pp.OneOrMore(pp.Word(pp.alphanums))
```
ptmcg, line 7552: `restOfLine` is probably more appropriate for `extraExpr`.

ptmcg:
To use the ignoreExprs code, try putting '|' delimiters between your columns, and then add `.ignore('|')` to make GoToColumn ignore them.

tests/test_unit.py
mattcarmody: Ah thank you! Good call, I don't want to introduce misleading examples.

mattcarmody:
Thanks for the feedback, I've updated the test to incorporate it! I recommend taking a look at the effect `.ignore('|')` is having on the ParseResults. The first `|` is being skipped and excluded from the results but not the second.

Issue #211, opened 2020-05-04
Getting an entire grammar as a string


TMiguelT:
I want to output my entire grammar in a vaguely human-readable format, so people can see what my parser is doing without having to necessarily understand pyparsing.

If I `print(str(grammar))` I get:
```
{{{LineStart Suppress:(<SP><TAB><CR><LF>)} flag_block} ^ {Suppress:(":") flag_block}}
```
This is a good start, but it's not recursively expanding tokens like `flag_block`. Is there any way to force this to happen in order to get a complete description of the grammar?

TMiguelT:
I think this is happening because I've explicitly `setName()` on `flag_block`. Is there a way to ignore the token name when converting to string?

ptmcg:
No, the purpose of setName is exactly to work as you describe. It helps make an expression like `Word(alphas, alphanums).setName("identifier")` to show up in exceptions as "Expected identifier" instead of "Expected W:(ABCD....)".  There have been other suggestions in the past also about displaying pyparsing parsers using "railroad track" diagrams, but even that effort would eventually look to some mechanism like setName that would support the definition of a friendly name for an expression, instead of a cryptic shorthand. Which items should be shown as names vs. their parsing constructs? Ultimately, you control that by which expressions you name using setName.

TMiguelT:
Well since I want the full grammar, I was hoping that every "composite" token (and, or, sequence) could be decomposed to their contents, but atomic components (literal, regex, anything that implements its own parse function) should stay as they are since their internal behaviour can't really be represented in a grammar. 

A railroad diagram would be super useful as well.

I guess from your answer, there is no mechanism to do this? What I might do is write a function that descends the grammar tree, runs `delattr(token, 'name')`, and then calls `str()` on the entire grammar which should mean that all tokens are decomposed wherever possible. Would this work?

ptmcg:
> What I might do is write a function that descends the grammar tree, runs delattr(token, 'name'), and then calls str() on the entire grammar which should mean that all tokens are decomposed wherever possible. Would this work?

Yes, I think that would work. You will get a long ugly mess though.

TMiguelT:
Hmm well railroad diagrams certainly would be ideal. If I wrote a pyparsing railroad generator for them using something [this library](https://github.com/tabatkins/railroad-diagrams/blob/gh-pages/README-py.md), do you think that should be something I PR into the core? Maybe behind an `extras` install such that you `pip install pyparsing[railroad]`?

[...]

PR #218, opened 2020-05-26
Railroad Diagrams

TMiguelT:
# Motivation
Refer to #211, in which I thought I wanted the grammar as a string, but realised I probably just wanted a railroad diagram. In general this solves the problem of "how to I visualise my grammar?"

# Summary
This produces railroad diagrams, along the lines of [json.org](https://www.json.org/json-en.html), for any arbitrary pyparsing grammar.

# Installation
For this module to work, you have to `pip install pyparsing[diagrams]` (replacing `pyparsing` with this git repo or the path to your local directory if you are developing). This will install the extra libraries needed to do this. They could be bundled into the main installation but this might not be ideal for users who want a lightweight installation.

# Usage
The public interface of this change consists of two functions: 

* `pyparsing.diagram.to_railroad()` which converts a PyParsing element tree into a diagram understood by the`railroad` library
* `pyparsing.diagram.railroad_to_html()` takes a list of such diagrams and converts them to HTML

These functions are documented fairly well in the code itself.

# Examples
[simpleBool.pdf](https://github.com/pyparsing/pyparsing/files/4681635/simpleBool.pdf) (from `examples.simpleBool`)
[json.pdf](https://github.com/pyparsing/pyparsing/files/4681634/json.pdf) (from `examples.jsonParser`)

# Tests

* Currently I have two tests, which involve running the above grammars through my diagram generator. More could be added
* By default, the test deletes the resulting HTML as soon as it's generated, to avoid cluttering up `/tmp`, however if you `export RAILROAD_DEBUG=1`, the files will not be deleted, and the tests will print out the filenames for the HTML files so you can manually inspect them.

# Potential Issues
* The HTML looks pretty bad. This is because I'm currently writing out an absolutely minimalist HTML document. This could be solved either by adding a default stylesheet, or by adding classes to the generated HTML and allowing the user to provide custom CSS that works with these classes
* Some grammars end up being extremely wide (like the JSON example attached). This is a very tricky problem. The railroad library does support wrapping the diagrams downwards, but you need to explicitly choose when to do so. Thus, if we wanted to fix this, we'd need to do something like keeping track of the amount of horizontal space used at any given time, and wrapping once we hit a threshold. Alternatively we could just do nothing and accept that no-one prints stuff anymore so they can just scroll across on the resulting webpage as they need.
* The complexity of certain elements is lost. For example, `MatchFirst` is displayed the same as `Or`. This is unfortunate but I see no easy solution. Ultimately the goal is to provide a readable image which trumps the possible desire to describe the grammar faithfully
* The variable names used within a grammar are lost if the user doesn't use `.setName()`. For example, we have token names like `jsonObject`, `jsonArray` etc in `examples.jsonParser`. In this case I think it's up to the user to remember to use `.setName()`, which isn't all that hard to do.


ptmcg:
Wow, the examples look great! I'll dig into this a little deeper, but I think you have done a great job here.

TMiguelT:
Great! Let me know if you need any changes.

ptmcg:
I'm going to merge this as-is, I'll add a blurb to the CHANGES file. Thanks again! 

After I push out this release, you should post some "look what I did" posts on blog, reddit, hackernews, etc.

Would really be great to add an example/demo somewhere and/or docs too. Doesn't have to be too detailed, judging from your test code your API is pretty simple.

TMiguelT:
Oh. Okay, let me add some docs and maybe some more tests before you release. I'll put in another PR hopefully soon.

PR #219, opened 2020-06-02
Additional configuration for `skipWhitespace` and `leaveWhitespace`

### LINKED ISSUE #203: Parent element's `skipWhitespace` seems to determine whether to skip whitespace
## Background
I have a segment of my grammar designed to match all the words on one line, ignoring initial whitespace. The grammar is:
```python
from pyparsing import *

tok = LineStart() + OneOrMore(Word(alphas))
test_string = '''
               some text on a line
          : more text on a line
'''
```

Now, if I try to actually match some text, this won't match because the first character of each line is whitespace, so `LineStart()` can't match:
```python
print(tok.searchString(test_string))
```

I can try to use `leaveWhitespace()` on the `LineStart()`, which seems to solve the issue. This is a bit strange considering `leaveWhitespace()` is "mostly used internally to the pyparsing module, rarely used by client code", but whatever, it works:
```python
tok = LineStart().leaveWhitespace() + OneOrMore(Word(alphas))
print(tok.searchString(test_string)[0])
```

## Issue

But now, I need to compose `tok` using an Or statement. This doesn't match the first line at all, even though it should, because `grammar` now has `skipWhitespace` set to True, even though `tok_a` has it as `False`:
```python
tok_a = LineStart().leaveWhitespace() + OneOrMore(Word(alphas))
tok_b = ':' + OneOrMore(Word(alphas))
grammar = tok_a ^ tok_b
grammar.searchString(test_string)
```

Alternatively if I set `grammar.leaveWhitespace()`, all the `Word()` tokens will fail to match because their parent element isn't skipping whitespace anymore, and the parent element, not the child, seems to determine whether they skip whitespace or not: 
```python
grammar.leaveWhitespace().searchString(test_string)
```

So in general, I'm finding it tricky to work around this behaviour whereby the parent elements are actually determining whether or not to skip whitespace.

TMiguelT:
* Add the `ignoreWhitespace()` method, which is the opposite of `leaveWhitespace()` (it sets `skipWhitespace=True`
* Add the argument `recursive=True` to both the above methods, to allow you to control which elements this affects

This closes #203 and helps with some other annoying issues I've been facing.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1343,6 +1343,14 @@ class ParserElement:
         """
         return Suppress(self)
 
+    def ignoreWhitespace(self):
+        """
+        Enables the skipping of whitespace before matching the characters in the
+        :class:`ParserElement`'s defined pattern.
+        """
+        self.skipWhitespace = True
+        return self
+
     def leaveWhitespace(self):
         """
         Disables the skipping of whitespace before matching the characters in the
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1343,19 +1343,23 @@ class ParserElement:
         """
         return Suppress(self)
 
-    def ignoreWhitespace(self):
+    def ignoreWhitespace(self, recursive=True):
         """
         Enables the skipping of whitespace before matching the characters in the
         :class:`ParserElement`'s defined pattern.
+
+        :param recursive: If true (the default), also enable whitespace skipping in child elements (if any)
         """
         self.skipWhitespace = True
         return self
 
-    def leaveWhitespace(self):
+    def leaveWhitespace(self, recursive=True):
         """
         Disables the skipping of whitespace before matching the characters in the
         :class:`ParserElement`'s defined pattern.  This is normally only used internally by
         the pyparsing module, but may be needed in some whitespace-sensitive grammars.
+
+        :param recursive: If true (the default), also disable whitespace skipping in child elements (if any)
         """
         self.skipWhitespace = False
         return self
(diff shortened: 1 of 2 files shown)


ptmcg:
Thanks again - do you want me to use your GitHub handle or full name in the acknowledgment in the CHANGES blurb?

TMiguelT:
Hi @ptmcg have you had a chance to look at this? My current project is somewhat blocked by this, so I'd love it to get merged. Let me know if you have any concerns about the changes.

ptmcg:
I need a little more convincing on this. If you want to explicitly skip whitespace on an expression that otherwise would not, you can prefix it with an Empty(). Your test cases use Literals with leading whitespace, which is generally bad pyparsing practice anyway (though I realize that sometimes test cases have to do such things in order to exercise corner cases).

LineStart() used to be kind of fuzzy and hence unpredictable, so I tightened it up to pretty much match only at column 1: at the beginning of the string or right after a newline. To work properly, you have to set default whitespace chars to leave out '\n', and then explicitly parse for them. Given your experience with pyparsing, I expect you have already done these things, but I'm just double checking.

And were you looking for this in a 2.4.x release or 3.0.0 release?

TMiguelT:
Well, I'm parsing a type of file which is highly whitespace-dependent, but only in certain parts. In other parts, it would become painful to separate each and every token using `White()` (I could do it, though). In general I realise it's an uncommon use-case, but I figured if `leaveWhitespace()` was already in there, you probably wouldn't mind including its inverse.

I didn't know about the `Empty()` trick. Does that work for skipping whitespace in between tokens as well as at the start of the whole expression? In any case, isn't using `Empty` in this way going to result in a more confusing grammar than just two opposite helper functions?

I was hoping for `2.X` since I've tried to ensure this isn't a breaking change, but if you consider it too large a change for `2.X` I suppose `3.X` would be fine.

ptmcg:
Revisiting this now  - as part of the discussion on #203 - some of these decisions date pretty far back to early pyparsing development days. What convinced me was the presence of the `skipWhitespace` attribute, which can be externally set to explicitly turn whitespace skipping on or off, and since I did not prefix it with '_', it is intended to be in the public API. I think your `ignoreWhitespace(recurse:bool)` is a better interface for doing this work, since it also takes care of the likely recursive case.

As for 2.4.8, I'm holding back on releasing it until 3.0.0 is at least up to rc1, since I want it to be the last release to include any backported features. But this will probably be included in it.

TMiguelT:
Great, thanks for the merge! For future reference, if you ever add that `recurse()` functionality for walking the element tree, I could clean up some of these methods. I think I'd only have to define each of them once.

Issue #223, opened 2020-06-19
Storing the default names of elements


TMiguelT:
So currently elements have their name stored in the `name` field, which is assigned to elements in their `__init__()`, but can then be overwritten with `setName()`.

The problem with this approach is that we then lose access to the original name, which can be useful as a description of what the element parser. In particular I want access to this for the railroad diagram generator.

So I'm proposing a change that allows us to keep the original names. Here's my current suggestion:
```python
from abc import ABC

class Token(ABC):
    # Other methods
    def __init__():
        self.name = None

    def setName(name):
        self.name = name
    
    def __str__():
        return self.name or self.defaultName

    @abc.abstractmethod
    @property
    def defaultName():
        pass

class Literal(Token):
    def __init__(self, matchString):
        super().__init__()
        self.match = matchString

    def defaultName():
        return '"%s"' % str(self.match)
```

This way, I can use `element.defaultName` in my diagram generator. Thoughts?

ptmcg:
I think the idea is fine, but I would make `name` the property, looking at `self._name or self.default_name`, and `default_name` be just an ordinary attribute, and the `name` setter would assign to `self._name`.

ptmcg:
You got me looking at these str and repr methods, and I found a huge amount of copy/paste code (about 100 lines!) for different implementations of `__str__` in the various classes. I've streamlined them now (and in the process, made the Word class emit a better looking string). I think you can define a property using `self.name if self.name is not None else self.strRepr`. `defaultName` is okay, but what do you think of calling it `docName`, as it is useful as the documentation name for an expression (similar to 'docstring')? But either name works for me.

In the course of defaulting `self.name` to None in `ParserElement.__init__`, I have broken your railroad code, so we are currently failing TravisCI tests. I think you will end up replacing direct references to name to references using the new property.

TMiguelT:
Well, nice job cleaning that up! I'm happy to call it `docName`.

>  This property won't need to be abstract, since all the subclasses should set name and/or strRepr appropriately
I would have thought this means it *should* be an abstract method, since that forces child classes to implement it

> In the course of defaulting self.name to None in ParserElement.__init__, I have broken your railroad code, so we are currently failing TravisCI tests.
That's fine, I'm rewriting it anyway since I worked out some massive flaws in it, so I'll just merge in master and fix it all in one go.

ptmcg:
[...]

PR #225, opened 2020-06-23
Diagram improvements IV

TMiguelT:
## Changes:

* Skip a lot more "empty" groups in the diagram that don't add much to the output
* Fix for elements with `name=None`, with a new test
* Add some assertions to the existing tests to fix major breaking changes
* Sort all elements by order found, which should help readability
* Use new `element.recurse()` method for listing children
* Use the new naming API discussed in #223 to great effect

## Example
The new diagram for the SQL grammar: 
[SQLRailroadv3.pdf](https://github.com/pyparsing/pyparsing/files/4827084/SQLRailroadv3.pdf)

## Blocked by
* [x] Resolution of https://github.com/pyparsing/pyparsing/issues/223#issuecomment-647950809. It's currently unclear how I can access *only user-defined* names, but not auto-generated names, and vice-versa. This is required for some final changes.

## Todos
* [x] Update the documentation with the newest SQL example diagram before we merge
* [x] Uncomment [this stuff](https://github.com/pyparsing/pyparsing/pull/225/files#diff-6eaf53d6899abb5f5d821bf06bc048e1R377-R387) once the above is resolved.


ptmcg:
`name` attribute will contain user-defined names; `strRepr` will contain the internally-generated names; `str()` will return `name` if not None, else `strRepr`. Thanks for the fix on `NotImplemented` vs `NotImplementedError`.

TMiguelT:
`name` will contain user-defined names, or it is supposed to now? Because currently it doesn't:

* https://github.com/pyparsing/pyparsing/blob/master/pyparsing/core.py#L1739
* https://github.com/pyparsing/pyparsing/blob/master/pyparsing/core.py#L1788
* https://github.com/pyparsing/pyparsing/blob/master/pyparsing/core.py#L1799
* https://github.com/pyparsing/pyparsing/blob/master/pyparsing/core.py#L1836

etc

ptmcg:
Hmmm, this is a bigger issue than I gave it credit. I am composing an email with more details so we don't clutter this PR with back-and-forth comments, and then we can summarize the plan in this PR once we settle on one.

TMiguelT:
Feel free to put all your thoughts into #223, since that's really what the issue is for.

TMiguelT:
Okay this is now ready, thanks for the naming API fixes in #223!

Although I'm happy for this to be merged as-is, I honestly wouldn't mind a quick eye over the actual code at some point. It needn't be now, but before this is released into a production version.

For example, I have some "simplifying" code that basically prunes the token tree in order to avoid processing certain elements that clog up the diagram. I entirely skip past[`Group` and `Forward` that themselves don't have a custom name](https://github.com/TMiguelT/pyparsing/blob/diagram-improvements/pyparsing/diagram/__init__.py#L300) and I extract [`Empty` elements that don't have a `customName`](https://github.com/TMiguelT/pyparsing/blob/diagram-improvements/pyparsing/diagram/__init__.py#L355-L357). Are there any other elements that fit into these two categories that I should be including? Or is there perhaps an edge case I'm forgetting by doing this?

TMiguelT:
Oh and perhaps another point for discussion: I'm abusing the "group" element in the railroad library a fair bit (which adds these dotted lines around things) 
image.

However I use these to label `customNames` as well as `ParseElementEnhance` as well as some `Groups`, so I'm open to suggestions to a more consistent way to label all these things.

ptmcg:
I'm going to merge as-is, we are still in alpha pre-releases so there is room for refinements. I'll need to give this some runthroughs with some parsers to get a feel for these questions.

ptmcg:
I think another type you might want to suppress would be `FollowedBy`, or at least the internal `_FB` subclass used in `infixNotation`.

I'm also happy to see the new Word default strings in these diagrams.

TMiguelT:
> I think another type you might want to suppress would be FollowedBy, or at least the internal _FB subclass used in infixNotation

Agreed, I'll try to add this in a future update

> I'm also happy to see the new Word default strings in these diagrams.

Indeed! Part of the naming changes I requested in #223 were to allow me to still show the accepted characters of terminals even if they have a custom name

PR #228, opened 2020-07-08
The metod getTokensEndLoc no longer exists

JohnnyVM:
Remove the metod getTokensEndLoc from documentation because no longer exists


ptmcg:
Nice catch, thank you!

pablo-angulo:
BTW, can you provide anywhere an explanation why this is no longer needed, and what is now the right way to do that job?
I know this happened a long time ago:
https://github.com/sagemath/sagetex/issues/6
but I would appreciate an alternative.

pablo-angulo:
...and the old code for `getTokensEndLoc` still works.

PR #229, opened 2020-07-18
Fixing generated default name for QuotedString

joshcoales:
Looks like an issue in changing the type of string format used


ptmcg:
Nice catch, thanks!

PR #233, opened 2020-08-14
Remove identChars override from Keyword.copy

jgrey4296:
The current behaviour of Keyword.copy unintuitively overrides the identChars of the copied keyword back to the DEFAULT_KEYWORD_CHARS. 

As such:
`
a_keyword = pp.Keyword("test", identChars="_") 
b_keyword = a_keyword.copy() 
`
Match different patterns.

This appears to be either a mistake, or is undocumented.
This pull request removes that override and adds a test to verify.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1992,7 +1992,6 @@ class Keyword(Token):
 
     def copy(self):
         c = super().copy()
-        c.identChars = Keyword.DEFAULT_KEYWORD_CHARS
         return c
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


ptmcg:
Oddly, this code has been this way ever since the Keyword class was introduced. I don't have good notes going back that far to remember why I felt that the copy should reset its `identChars`.

It looks like this is a good change (with added unit test too - thanks!), but since you are removing what is really the only purpose for this override of `copy()` from the subclass, go ahead and just delete `Keyword.copy()` entirely.

I'll add a blurb to the CHANGES file after merging your PR.

ptmcg:
I'll take care of the cleanup from here, as I'd like to move forward with some other changes. Thanks again for submitting!

PR #236, opened 2020-08-17
Some proposed changes to pp_diagram output, plus monkeypatch create_diagram method onto ParserElement

ptmcg:
@TMiguelT - I've had these changes for a while, but wasn't sure about the Git incantations to push them properly. I think this is... oh, crap, I should have done this in a branch. Well, I can just nuke this forked repo when we are done, rather than try to patch it up.

These changes are not complete, there are still some issues with some of the labeling I think, and proper referencing to a named element if it is used multiple times in a high expression.

I also would like to add the results name to the dashed border now, since the expression name is getting written on the block element itself. If you can point me to where this is set, I'll take a run at it.

What do you think of monkeypatching `.create_diagram` onto `ParserElement`?


pyparsing/diagram/__init__.py
```diff
@@ -271,6 +278,12 @@ def _worth_extracting(element: pyparsing.ParserElement) -> bool:
     )
 
 
+def _already_extracted(
+    element: pyparsing.ParserElement, lookup: ConverterState
+) -> bool:
+    return any(element.name == diag.kwargs["name"] for diag in lookup.diagrams.values())
```
TMiguelT, line 284: The `ConverterState` object indexes both of its dictionaries using the pyparsing element ID. So you can just check `return id(element) in lookup.diagrams` here.

pyparsing/diagram/__init__.py
```diff
@@ -341,9 +354,9 @@ def _to_diagram_element(
             ret = EditablePartial.from_call(railroad.Sequence, items=[])
     elif isinstance(element, (pyparsing.Or, pyparsing.MatchFirst)):
         if _should_vertical(vertical, len(exprs)):
-            ret = EditablePartial.from_call(railroad.HorizontalChoice, items=[])
```
TMiguelT, line 344: Good catch, I evidently had this wrong. Let me know if it actually looks better this way, though.

pyparsing/diagram/__init__.py
```diff
@@ -355,6 +368,15 @@ def _to_diagram_element(
     elif isinstance(element, pyparsing.Empty) and not element.customName:
         # Skip unnamed "Empty" elements
         ret = None
+    elif _is_named_terminal(element):
+        if not _already_extracted(element, lookup):
+            ret = EditablePartial.from_call(
+                railroad.Terminal, title=element.name, text=element.defaultName
+            )
+        else:
+            ret = EditablePartial.from_call(
```
TMiguelT, line 377: I think this check (`if not _already_extracted`) could be simplified just by adding a new condition to `_worth_extracting`. As it stands, we already check if it's `_worth_extracting`, and if it has already been extracted, then I already have code for creating a `NonTerminal` here: https://github.com/pyparsing/pyparsing/blob/1b86c4915003ed6bd9d47126f9337801be3853b0/pyparsing/diagram/__init__.py#L332-L345.

So if you simply add `if _is_named_terminal(element): return True` to `_already_extracted`, I think you get this behaviour for free, with less and simpler code.

pyparsing/diagram/__init__.py
```diff
@@ -355,6 +368,15 @@ def _to_diagram_element(
...
+                railroad.NonTerminal, title=element.name, text=element.name
```
TMiguelT, line 378: `title` will only set the hover text, I assume you know that? This will make it harder to see both the `customName` and the `defaultName` compared to my current solution of using a group to do that: https://github.com/pyparsing/pyparsing/blob/1b86c4915003ed6bd9d47126f9337801be3853b0/pyparsing/diagram/__init__.py#L387-L393

pyparsing/diagram/__init__.py
```diff
@@ -379,6 +401,12 @@ def _to_diagram_element(
         number=lookup.generate_index(),
     )
 
+    if _is_named_terminal(element):
```
TMiguelT, line 404: So here you're bypassing the normal mechanism of "only extract if we've seen it twice". Is this intentional? I forget what we discussed. If so, there's probably an easier way to do this than using the extraction mechanism at all, but I can't think of exactly what it might be.

pyparsing/diagram/__init__.py
```diff
@@ -424,3 +452,27 @@ def _to_diagram_element(
         )
     else:
         return ret
+
+
+# monkeypatch .create_diagram method onto ParserElement
+def _create_diagram(expr: pyparsing.ParserElement, output_html: Union[str, TextIO]):
```
TMiguelT, line 458: I think this might be better as a method on the original `ParserElement` class, which checks for an `ImportError` as you have done at the top. I prefer this to monkey patching, anyway.

ptmcg: My intention behind monkeypatching is so that the core library can be easily vendored without having to include dependencies on other packages beside the stdlib.

pyparsing/diagram/__init__.py
ptmcg: Yeah, I think this is intentional. Even if an object is only seen once, if it is sufficiently complex (and has a name), it should be a separate subdiagram. This makes the diagrams more likely to map to a BNF.

pyparsing/diagram/__init__.py
ptmcg: The default name gets displayed in the element's subdiagram, so we shouldn't need to show both defaultName and customName at the same time.

pyparsing/diagram/__init__.py
ptmcg: Ok, much simpler!

pyparsing/diagram/__init__.py
TMiguelT: Oh, okay. I think I need to see an example of this.

pyparsing/diagram/__init__.py
TMiguelT: But if you do this, is that not sufficient? Then the dependencies are still optional.
```python
try:
    import pyparsing.diagram
except ImportError:
    raise Exception('You need to run `pip install pyparsing[diagrams]`')

ptmcg: I got rid of the monkeypatch and added create_diagram to ParserElement as you suggested.

pyparsing/diagram/__init__.py
ptmcg: Was folded into master already.

ptmcg:
Closing, as all these changes have been either rejected or incorporated into the latest 3.0.0rc1 release.

PR #256, opened 2020-12-27
Add Python 3.9 support

eumiro:
`py39-dev` has been removed in 5811c79597aa718df2e72fb95189233055a2ded3, now the final Python 3.9 version can be used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ matrix:
       env: TOXENV=py37
     - python: 3.8
       env: TOXENV=py38
+    - python: 3.9
+      env: TOXENV=py39
     - python: pypy3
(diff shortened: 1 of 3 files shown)


ptmcg:
Thanks for submitting this!

eumiro:
Thank you @ptmcg , would you consider dropping Python 3.5 support as well (now after EOL)?

ptmcg:
I'm going to do this in pyparsing 3.1. I don't expect f-strings will be a significant change to the core library, but more relevant to the examples. The bigger impact of dropping 3.5 will be the addition of types to the public method signatures, and that will be a significant job, more than I want to take on in this long-long-long-overdue release.

PR #258, opened 2021-01-03
Update whats_new_in_3_0_0.rst

retsyo:
since `pip install railroad` install anthor package, I think it is better to point out the corrected one.


ptmcg:
Of course! Would you also update the exception message in core.py?

retsyo:
> Of course! Would you also update the exception message in core.py?

done

PR #280, opened 2021-05-14
Fix misc. documentation typos

luzpaz:
Found via `codespell -q 3 -L ba,fourty,halp,inout,strng`


ptmcg:
These look great, thanks a bunch!

PR #281, opened 2021-05-16
add a caseless parameter to the CloseMatch class

### LINKED ISSUE #279: Add a Caseless version of CloseMatch
I notice that theres classes for `CaselessLiteral` and a `CloseMatch` but there doesnt seem to be a class combining the two for situations where you want a `CaselessLiteral` but also want to be tolerant of a certain number of incorrect characters (i.e. to catch typos) as is the behavior with `CloseMatch`. Whis would have the behavior of a less-stringent `CloseMatch` that will only consider a character to be nonmatching if it is a different character, rather than just a differently-cased version of the same character.

After poking around the code, it would seem to me that this could essentially be some light refeatoring to move the string comparison from [here](https://github.com/pyparsing/pyparsing/blob/b5b1fdd1c0a3dfffbbad206e885d59e6015015a8/pyparsing/core.py#L2154) to a new function within `CloseMatch` and then create a subclass of `CloseMatch` that changes the comparison function to use `src.lower() != mat.lower()) as the comparison instead.

Would this be something that would be useful to submit a pull request for? or would this be better to create in a separate package/module?

MoralCode:
fixes #279 

in addition to unit tests, this change also includes formatting changes to unrelated code that the pre-commit hook required me to do before committing.

My machine says that tests are failing, however it seems to be due to tests that are unrelated to the new changes or new tests that I have added. These tests were failing when i ran the commands to set up the testing environment as suggested in the `tests/README` file and i'm not sure how to properly deal with them since this is my first contribution to this codebase

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -56,6 +56,9 @@ Version 3.0.0c1 -
 - Fixed bug in QuotedString class when the escaped quote string is not a
   repeated character. (Issue #263)
 
+- Added a caseless parameter to the `CloseMatch` class to allow for casing to be ignored when checking for close matches
+
+
 
 Version 3.0.0b3 - August, 2021
 ------------------------------
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -2380,6 +2380,7 @@ class CloseMatch(Token):
     :class:`CloseMatch` takes parameters:
 
      - ``match_string`` - string to be matched
+     - ``caseless`` - a boolean indicating whether to ignore casing when comparing characters
      - ``max_mismatches`` - (``default=1``) maximum number of
        mismatches allowed to count as a match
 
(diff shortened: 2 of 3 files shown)


MoralCode:
@ptmcg is this something that you think aligns well with the goals of pyparsing? happy to do my best to fix the merge conflicts if this would be a helpful merge

ptmcg:
My apologies - I just reviewed the issue you posted, and you submitted this PR around May!

Since then there have been quite a number of commits. Try rebasing your code off the latest master, or if that fails, probably just start a new PR with a fresh branch of latest master. I am concerned that resolving merge conflicts will leave out needed changes (this just happened to me last week at work!).  I'll work with your PR as soon as you let me know its ready.

MoralCode:
@ptmcg just did those updates. looks like most of it was due to me committing autoformatting changes from linters which seems to have been why there were so many conflicts.

The only question I have is if I put the entry to CHANGES in the right spot (3.3.3c1)

ptmcg:
I think it was good, if I have to move it, I'll move it.

ptmcg:
Released in 3.0.0rc1

Issue #284, opened 2021-05-31
[Feature request] Add new flags to `delimitedList` helper.


heckad:
1) Allow last delimiter flag with type bool(default false). If true then should parse `delimitedList(arg) + Optional(Suppress(','))`.
2) Allow empty. Equeal to `Optional(delimitedList(arg), [])`

ptmcg:
For point 1, I had a similar idea recently, and I thought I had already added it in, as something like `allowTrailing=True`. Yes, I think this is a very good idea, for parsing syntaxes that allow the extra trailing comma (like Python).

For point 2, I'm less convinced that another flag to learn is better than using Optional. And if Optional, it shouldn't have a default of '[]', but just, well, nothing. If it needs to be a list, then it would be wrapped in a Group, and the Group would take care of making it a list.

So look for your first suggestion in the next pyparsing release. Thanks for writing, and welcome to pyparsing!

heckad:
1) Please do this as [kw args only](https://www.python.org/dev/peps/pep-3102/).

2) This is needs for a function call. If parse something like `foo()` then call args will be empty list `[]`.

Thank you! 

PS Can I create the pull request?

ptmcg:
Again, I don't think you want 2 to default to []. If your function args are Grouped, then the Group will create an empty list for you. If they aren't Grouped, then you'll just get nothing, in place of any.

But go ahead and do the PR for the first item, `allowTrailing`. (My apologies for the camelcase names, but that is how just about all the other kwargs are done in this version - even PEP 8 values consistency in this regard.)

Keyword-args only is fine for new args, so okay for this new kwarg, but breaks for old args. I like the keyword-args only form, but I'll probably only add it throughout when I break all the APIs by making them PEP 8 snake case names.

ptmcg:
allow_trailing_delim flag added in 3.0.0b3

PR #285, opened 2021-06-08
Add allowTrailingDelim to delimitedList helper

heckad:
https://github.com/pyparsing/pyparsing/issues/284

### DIFF
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -6,7 +6,7 @@ from .util import _bslash, _flatten, _escapeRegexRangeChars
 #
 # global helpers
 #
-def delimitedList(expr, delim=",", combine=False):
+def delimitedList(expr, delim=",", combine=False, *, allowTrailingDelim=False):
     """Helper to define a delimited list of expressions - the delimiter
     defaults to ','. By default, the list elements and delimiters can
     have intervening whitespace, and comments, but this can be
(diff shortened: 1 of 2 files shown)


pyparsing/helpers.py
```diff
@@ -22,10 +22,13 @@ def delimitedList(expr, delim=",", combine=False):
         delimitedList(Word(hexnums), delim=':', combine=True).parseString("AA:BB:CC:DD:EE") # -> ['AA:BB:CC:DD:EE']
     """
     dlName = str(expr) + " [" + str(delim) + " " + str(expr) + "]..."
+
+    trailing_delim = Optional(delim) if allowTrailingDelim else Empty()
```
ptmcg, line 26: Adding a trailing `Empty()` can do unintended whitespace skipping at the end of the expression when no trailing delimiter is used. Need to reimplement so that without `allowTrailingDelim=True`, you only get the expression as it currently is defined.  Probably refactor to something like:

    if not combine:
        delim = delim.suppress()
    
    delimited_list_expr = expr + (delim + expr)[...]
    if allowTrailingDelim:
        delimited_list_expr += Optional(delim)
    
    if combine:
        return Combine(delimited_list_expr)
    else:
        return delimited_list_expr

ptmcg:
Note that the string used in setName() is not 100% accurate if allowTrailingDelim is True. Should probably update that also.

heckad:
@ptmcg?

ptmcg:
Sorry for being off-line for so long. Still learning my way around at the new job, and current project is pretty urgent.

I'll move your PR to the front of the line.

Issue #267, opened 2021-02-03
Provide a way to disable packrat caching


arekfu:
I would like to have a `ParserElement.disablePackrat()` method to disable packrat caching. Different packages importing pyparsing within the same application may have different opinions about the usefulness of packrat caching. Currently, if any one of them calls `enablePackrat()`, all the other packages are stuck with it, with large performance losses in some cases.

The ideal solution would be to make it possible to choose whether to use packrat caching on a parser-by-parser basis, and not globally for the whole `pyparsing` package, but I understand that this may take a lot of refactoring.

ptmcg:
Sadly, yes, this would be a big refactoring project.

You *can* pass an argument to `enablePackrat` to set the size of the cache (the default is 128), `None` to set an unbounded cache, or `0` to set effectively *no* cache. This might work for you.

arekfu:
Thanks for looking into this. Looking at the code though it seems that setting `cache_size_limit=0` is not quite the same as having never called `enablePackrat()` in the first place. Maybe we could special-case `cache_size_limit=0` to set `self._parse` back to `self._parseNoCache`?

maxfischer2781:
Since packrat does not play well with recursion (see #287), I've also pondered adding an off-switch for both. Would it be too much feature creep if I put such a switch in the recursion PR?

ptmcg:
Not at all, it seems very closely related.

ptmcg:
With the addition of left recursion support, there is now the new method `ParserElement.disable_memoization()`, released in 3.0.0b3

PR #288, opened 2021-06-19
Left-Recursion support

### LINKED ISSUE #287: Support direct and indirect left-recursion (Medeiros et al.)
PyParsing currently does not support "left-recursion", i.e. recursion without advancing the parsing position. Since left-recursion is the natural way to express certain common grammatical features, such as left-associativity, many grammars require cumbersome workarounds.  
As PyParsing is practically a PEG parser, some approaches from recent literature on left-recursive PEG should be compatible. Notably, since PEG itself is not left-recursive, the exact behaviour of left-recursion depends on the specific approach.

As suggested in ["Left Recursion in Parsing Expression Grammars" by Medeiros, Mascarenhas, Ierusalimschy](https://arxiv.org/abs/1207.0443), left-recursion can be modelled as bounded recursion. This is roughly equivalent to duplicating *variable* clauses into a ladder until the longest match is achieved:
```
# left-recursive PEG definition
E <- E '+' num / num
# bounded recursion PEG definition
E0 <- fail
E1 <- E0 '+' num / num
E2 <- E1 '+' num / num
E3 <- E2 '+' num / num
....
```
The variable clause seems to match PyParsing's ``Forward``. Changing ``Forward.parseImpl`` as described in the paper should allow direct and indirect left-recursion.

---------------

The approach by Medeiros et al. is the basis for [PEP 617: Left recursion](https://www.python.org/dev/peps/pep-0617/#left-recursion). The PEP uses an optimised but complicated solution that relies on re-using a Packrat cache. In order to work with and without ``enablePackrat``, to provide a simple base implementation, and speed up development, a direct implementation seems prudent.

maxfischer2781:
This PR adds support for direct and indirect left-recursion, according to the "bounded left-recursion" scheme. This is similar to but neither identical nor directly compatible with Packrat.

* [x] basic LR implementation
* [x] LR must be enabled like Packrat
* [x] tests for direct left recursion
* [x] tests for indirect left recursion
* [x] tests for non-PEG clause interactions
* [x] docs
* [x] ? non-action left recursion lookahead ?

Closes #287.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -773,6 +773,32 @@ class ParserElement(ABC):
         ParserElement.recursion_memos.clear()
 
     _packratEnabled = False
+    _bounded_recursion_enabled = False
+
+    @staticmethod
+    def enable_bounded_recursion():
+        """
+        Enables "bounded recursion" parsing, which allows for both direct and indirect
+        left-recursion.
+
+        Example::
+
+            import pyparsing as pp
+            pp.ParserElement.enable_bounded_recursion()
+
+            E = pp.Forward("E")
+            num = pp.Word(pp.nums)
+            E <<= E + '+' - num | num
+
+            print(E.parseString("1+2+3"))
+
+
+        Bounded Recursion parsing works similar but not identical to Packrat parsing,
+        thus the two cannot be used together.
+        """
+        if ParserElement._packratEnabled:
+            raise RuntimeError("Packrat and Bounded Recursion are not compatible")
+        ParserElement._bounded_recursion_enabled = True
 
     @staticmethod
     def enablePackrat(cache_size_limit=128):
@@ -4395,8 +4421,10 @@ class Forward(ParseElementEnhance):
                 "Forward expression was never assigned a value, will not parse any input",
                 stacklevel=stacklevel,
             )
-        # return super().parseImpl(instring, loc, doActions)
-        return self.parse_recursive(instring, loc, doActions)
+        if not ParserElement._bounded_recursion_enabled:
+            return super().parseImpl(instring, loc, doActions)
+        else:
+            return self.parse_recursive(instring, loc, doActions)
 
     def parse_recursive(self, instring, loc, doActions=True):
         with ParserElement.recursion_lock:
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -4463,8 +4463,6 @@ class Forward(ParseElementEnhance):
                         new_loc, new_result = prev_loc, prev_result
                     # the match did not get better: we are done
                     if new_loc == prev_loc:
-                        if isinstance(prev_result, Exception):
-                            raise prev_result
                         return new_loc, new_result
                     elif new_loc < prev_loc:
                         return prev_loc, prev_result
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -4460,7 +4460,7 @@ class Forward(ParseElementEnhance):
                 return prev_loc, prev_result.copy()
             except KeyError:
                 # we are searching for the best recursion expansion – keep on improving
-                # need to track both `doActions` cases separately here!
+                # both `doActions` cases must be tracked separately here!
                 prev_loc, prev_peek = memo[self, False] = loc - 1, ParseException(
                     instring, loc, "Forward recursion without base case", self
                 )
@@ -4477,15 +4477,19 @@ class Forward(ParseElementEnhance):
                     # the match did not get better: we are done
                     if new_loc <= prev_loc:
                         if doActions:
-                            _, prev_result = memo[self, True]
+                            # store the match for doActions=False as well,
+                            # in case the action did backtrack
+                            prev_loc, prev_result = memo[self, False] = memo[self, True]
                             return prev_loc, prev_result.copy()
                         return prev_loc, prev_peek.copy()
                     # the match did get better: see if we can improve further
                     else:
                         if doActions:
-                            # TODO: store errors
-                            # TODO: fail on backtrack? (we go out of sync otherwise)
-                            memo[self, True] = super().parseImpl(instring, loc, True)
+                            try:
+                                memo[self, True] = super().parseImpl(instring, loc, True)
+                            except ParseException as e:
+                                memo[self, False] = memo[self, True] = (new_loc, e)
+                                raise
                         prev_loc, prev_peek = memo[self, False] = new_loc, new_peek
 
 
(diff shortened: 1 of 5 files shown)


maxfischer2781:
The algorithm seems to interact very weirdly with ``Or``/``^``. The ``ParseResults`` seems to include *all* intermediate matches before the correct one:
```python3
import pyparsing as pp

pp.ParserElement.enable_bounded_recursion()

expr = pp.Forward('expr')
expr <<= (expr + "a") ^ (expr + "ab") ^ (expr + "abc") ^ "."

print(expr)
print(expr.parseString(".abca"))  # ['.', 'a', 'ab', 'abc', 'abc', 'a', 'a']

expr = pp.Forward('expr')
expr <<= (expr + "abc") | (expr + "ab") | (expr + "a") | "."

print(expr)
print(expr.parseString(".abca"))  # ['.', 'abc', 'a']
```
When there is no recursion in the clause the issue does not appear, even if recursion is switched on.

ptmcg:
I think we need this to go onto a different branch than master. Do you need me to create one?

maxfischer2781:
If you have a branch name in mind, feel free to create one.

ptmcg:
Created branch 'left_recursion_support'

ptmcg:
Could you add a test that is one of the classic LR examples?

    delimited_list_of_e := delimited_list_of_e ',' e | e

Or

    one_or_more_e := one_or_more_e e | e

Or we could package these in a left_recursion_examples.py in the examples dir.

maxfischer2781:
I've added a test for these two cases. Since the PR goes to a separate branch, reworking/recreating some examples can be done in that branch.

Can you think of anything else that should go in the PR for it to be self-contained?
- Is the naming fine with you? ``enable_bounded_recursion`` feels a bit long – would ``enable_recursion`` be okay?
- Any obvious cases that might need testing?
- Should the "disable all memos" from #267 be part of this PR or added later on?
- Running the Packrat tests with the Recursion parsing? I've only seen that there are some packrat specific checks to disable incompatible tests, can probably relax that to use the recursion parsing as well.
- I have an idea to restrict the Recursion memo – not quite as strict as the Packrat limit, but close. Should that go in directly or in a follow up PR?

I still have the PR in draft until I'm sure all vital parts are there, but the structure won't change significantly. So feel free to add comments on the code if you have the time already.

ptmcg:
Actually, I think I’d like you to be *more* explicit that you are enabling _left_ recursion, since recursion is already supported so it is already enabled 😊.

 

Can you handle “boundedness” in a manner similar to the way packrat will accept a limit, or default to unbounded?

 

Let’s not add much more to this PR, I’d like to merge it in and work with it a bit. Your comment about using copies of results in Or concerns me a little, with respect to memory and CPU utilization. Packrat parsing went through a huge refactor a few years ago to address performance and memory issues. And performance is generally a constant complaint about pyparsing. So I’ll be interested to see how adding LR support impacts things.

 

-- Paul

maxfischer2781:
> Can you handle “boundedness” in a manner similar to the way packrat will accept a limit, or default to unbounded?

I think so. The algorithm relies on having all currently expanding recursions in the memo, but it would be possible to free entries as soon as expansion for them is complete. So a soft limit would be possible, where the memo stores up to X entries that it already expanded, but might temporarily have more entries when it is expanding more than X entries at the same time.

> Your comment about using copies of results in Or concerns me a little, with respect to memory and CPU utilization. Packrat parsing went through a huge refactor a few years ago to address performance and memory issues. And performance is generally a constant complaint about pyparsing. So I’ll be interested to see how adding LR support impacts things.

This is something that might need testing/benchmarking. Since the copying and memoization only happens for the few ``Forward`` rules, it's not a straightforward "x times more operations/memory". For nontrivial well-behaved grammars, the algorithm should take less memory than Packrat (since it memoizes less) and be faster than the base parsing strategy (since it memoizes).
Though for grammars without backtracking and without left-recursion, there is no benefit and all the cost is wasted.

maxfischer2781:
Renamed the switch to ``enable_left_recursion``.

I've naively switched on the Packrat tests for left-recursion to see whether things work as expected. There are some issues from me being *too* naive (abusing the packrat flag to switch tests triggers some side-effects) but also some that look like genuine issues: matching too little, additional nesting, and keeping some values as ``str``s when they should be ``int``s.

maxfischer2781:
Some failures were due to the memo mixing the two ``doActions`` branches, causing the bare/transformed info to be returned wrongly. That was a trivial fix by adding ``doActions`` to the memo.  
However, this has messed up the ``Or`` clause again. Since ``Or`` repeatedly checks for the longest match with ``doActions=False``, one part of the memo grows very fast. Once it has made a decision and attempts to evaluate the single expression, it does so with the ``doActions=True`` memo that is still small.

Since the memo works only on  ``Forward`` elements, the execution count is not fixed/minimal. That's why ``testInfixNotationMinimalParseActionCalls`` fails – I think that's okay since (like Packrat) there is no strong guarantee on execution count with memoization.

I assume the issue with ``testIndentedBlockTest2`` is also memoization. What I don't get is that Packrat seems to be able to parse it. Any idea?

maxfischer2781:
I've tracked down all the failures to my recursion implementation not (yet😉) properly handling ``doActions``.
- For stack/side-effect problems, it runs actions too often because the action also run when it *tries* whether there is an improvement.
- For ``Or`` problems, the actions/no-actions branches diverge.

That's not a deficiency of the algorithm, but of me not using ``doActions`` well. The fix to both of these seems to be to run actions and no-actions in lockstep: use ``doActions=False`` to peek at whether the next iteration is better, then use ``doActions=True`` to get the actual result. Incidentally, that is similar to how ``Or`` works.
I'll set the PR ready for review – go ahead if you feel it's appropriate for the intermediate `left_recursion_support` branch already. I will add the fix as soon as I've figured out how to best translate it to code.

ptmcg:
This is excellent work! Looking forward to including it in the v3 release!

maxfischer2781:
Got it fixed, as per the strategy documented in the issue. The logic did not change much, though checking for `doActions`+`try:`/`except` to handle action failures bumps the indentation quite a lot.

The left-recursion now passes all tests of Packrat. 🎉 

I've got two more changes in the pipeline (memo size and memo disabling) that won't really affect how this works but are logically related - so they may-or-may-not end up in this PR, depending on how fast either of us manages it.

ptmcg:
Max - this code seems to have stabilized, is it ready to merge in its current state? I'll need to work with it to write up docs, etc.

maxfischer2781:
Yes, I'm pretty confident now that it passes the existing test suite. Good to go for merging from my side.

Do let me know if I can be of help with documenting etc., or if you find any rough edges that don't seem to work properly.

PR #295, opened 2021-08-12
Fix old method names in comparison

danielroseman:
Looks like these got over-zealously corrected to the new versions.


ptmcg:
Great catch, thanks!

Issue #226, opened 2020-06-25
Railroad diagram future updates


multimeric:
I'm making this issue so that I and anyone else can suggest improvements to the railroad diagram generation feature.

TODOs:
* [ ] (from #225) Suppress `FollowedBy`/`_FB` subclass
* [ ] Fix the root element not being called `Unnamed` and instead being called `Forward` e.g. in the SQL example
* [ ] Fix the root element not being labelled "root" as it was previously
* [ ] Consider ways to use the element class to disambiguate the Unnamed classes

ptmcg:
`expr = expr1 + expr2 + expr3` will create `expr` as `And([And([expr1, expr2]), expr3])`.  If the sub exprs are "collapsible" , calling `expr.streamline()` will flatten `expr` to `And([expr1, expr2, expr3])`.  This might help in verticalizing `And` elements. Probably something to add to `to_railroad`.

ptmcg:
I could see some value in including results names as some kind of annotation, perhaps as part of the label you attach in the dashed lines around the element. I could also see this as unwanted clutter to someone who just wants to see the structure of the parser. Perhaps a `show_results_names` option to `to_railroad`?

TMiguelT:
Thank you, that's useful to know about `streamline()`, although there is actually a chance that the railroad library streamlines nested `And`s anyway, because to my knowledge railroad diagrams don't have such a concept. 

`resultName` is tricky, because it is indeed useful, but as I said above, I'm already cramming so many things into the "group" annotations that it might get cluttered. I suppose I could have a consistent format like `tokenName → resultsName`. Do you think that should be in the annotation (above the token), or in the textbox itself?

ptmcg:
I just tried generating a diagram for the delta_time.py time_expression - very interesting. But it looks like streamline() doesn't make any difference.

On resultName, I too was thinking of something like a right arrow, but as I said, I could see some people wanting it and others not. So you could make it an argument to `to_railroad`, or just leave it out for now and see if anyone requests it. (And if `to_railroad` starts getting too many more args, it may be worth packaging as a Railroader object.

ptmcg:
What do you think about breaking out named terminals (like Word and Regex) as their own sub-diags. I think their usages would be cleaner-looking. See this paste HTML: https://dpaste.com/DHHV86L5J where I tried to modify your code to break them out, but only got as far as replacing the names.  This diagram is for this little parser:

```
ppc = pp.pyparsing_common
[...]

PR #298, opened 2021-09-05
226 railroad updates

ptmcg:


ptmcg:
I was premature with the dataclasses, since we go back to 3.6, and they weren't introduced until 3.8 >:(

multimeric:
> I was premature with the dataclasses, since we go back to 3.6, and they weren't introduced until 3.8 >:(

They were added in 3.7, it seems

pyparsing/diagram/__init__.py
```diff
@@ -102,26 +107,32 @@ def resolve_partial(partial: "EditablePartial[T]") -> T:
...
-        lookup.first[root_id].mark_for_extraction(root_id, lookup, force=True)
+    if root_id in lookup:
+        lookup[root_id].mark_for_extraction(root_id, lookup, force=True)
 
     # Now that we're finished, we can convert from intermediate structures into Railroad elements
-    resolved = [resolve_partial(partial) for partial in lookup.diagrams.values()]
+    diags = list(lookup.diagrams.values())
+    # collapse out duplicate diags with the same name
+    seen = set()
+    deduped_diags = [d for d in diags if d.name not in seen and (seen.add(d.name) or d)]
```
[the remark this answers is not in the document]
multimeric: Because `diagrams` is a dictionary indexed by element IDs. Is this not sufficient to enforce uniqueness?

pyparsing/diagram/__init__.py
```diff
@@ -200,17 +214,30 @@ class ConverterState:
...
+        self.unnamed_index: int = 1
         #: The index of the next element. This is used for sorting
-        self.index = 0  # type:  int
+        self.index: int = 0
         #: Shared kwargs that are used to customize the construction of diagrams
-        self.diagram_kwargs = diagram_kwargs  # type:  dict
+        self.diagram_kwargs: Dict = diagram_kwargs or {}
+        self.extracted_diagram_names: Set[str] = set()
+
+    def __setitem__(self, key, value):
```
multimeric, line 230: This is a fair change, since we mostly interact with this class's `first` attribute. Maybe add types for the params and return values here?

pyparsing/diagram/__init__.py
```diff
@@ -294,40 +325,64 @@ def _to_diagram_element(
...
-        )
+    if not element.customName:
+        if isinstance(element,
+                      (
+                          pyparsing.TokenConverter,
+                          pyparsing.Forward,
+                          pyparsing.Located,
+                      )
+                      ):
+            # However, if this element has a useful custom name, we can pass it on to the child
```
multimeric, line 344: "if this element has a useful custom name, and its child does not"

pyparsing/diagram/__init__.py
```diff
@@ -276,6 +305,8 @@ def _to_diagram_element(
     vertical: int = None,
     index: int = 0,
     name_hint: str = None,
+    group_results_name: str = None,
```
multimeric, line 308: Docstring for these params

pyparsing/diagram/__init__.py
```diff
@@ -294,40 +325,64 @@ def _to_diagram_element(
...
 
         elif el_id in lookup.diagrams:
             # If we have seen the element at least twice before, and have already extracted it into a subdiagram, we
             # just put in a marker element that refers to the sub-diagram
-            return EditablePartial.from_call(
+            ret = EditablePartial.from_call(
                 railroad.NonTerminal, text=lookup.diagrams[el_id].kwargs["name"]
             )
+            if show_results_names and element_results_name:
+                ret = EditablePartial.from_call(
```
multimeric, line 382: I'm confused by this. Aren't we already creating a group here: https://github.com/pyparsing/pyparsing/blob/97fcd3dbf6947d9df445357a499c99d5ede7e5a1/pyparsing/diagram/__init__.py#L481-L484? 

I guess this second code chunk is now no longer being called. But actually I would rather this behaviour is put there instead of here, inside the extraction logic.

tests/test_diagram.py
```diff
@@ -14,50 +14,58 @@ def railroad_debug(self) -> bool:
         """
         Returns True if we're in debug mode
         """
+        return True
```
multimeric, line 17: I assume this shouldn't have been committed?

ptmcg:
OK! I think I have resolved the infixNotation issue, and the Forward issues (at least in the context of infixNotation). Part of the issue was how I was doing setName()'s in infixNotation. The dependencies are subtle here, especially with recursion in generated parser code!

Need to rerun tests, but first results look promising.

tests/test_diagram.py
ptmcg: Gah, yes - I'll set this in my PyCharm config from now on.

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
ptmcg: id(element) is *too* unique. If an element shows up multiple times with different results names, pyparsing creates a copy of that element. This creates separate elements with separate id's, but the same name and same structure. Deduping by id still shows multiple diagrams for that same name/structure. For the purposes of diagramming, we don't need both copies, so deduping by name clears that up. It *does* put the burden on the parser developer to not duplicate names passed to setName(). And it poses an issue for a parser that might have more than one infixNotation, since that method auto-generates names for the internal recursive elements. Generally not a problem, but there is one name that could be a problem, "base_expression". It's a bug to work on during 3.0.0.c1

pyparsing/diagram/__init__.py
ptmcg: The name "first" just generally confused me, so getitem/setitem kept the logic and made the calling code more natural. Adding type annotations in next update.

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
multimeric: Oh I see. Well if you're using the name as a unique ID then you might as well change the `diagrams` dictionary to use them as indices. Part of the point of this dictionary is deduplication anyway. Or can we `hash` your elements or some other way that is less likely to collide than using names but still compares by value?

pyparsing/diagram/__init__.py
```diff
@@ -102,26 +107,32 @@ def resolve_partial(partial: "EditablePartial[T]") -> T:
...
     Convert a pyparsing element tree into a list of diagrams. This is the recommended entrypoint to diagram
     creation if you want to access the Railroad tree before it is converted to HTML
+    :param element:
     :param diagram_kwargs: kwargs to pass to the Diagram() constructor
     :param vertical: (optional)
     """
     # Convert the whole tree underneath the root
-    lookup = ConverterState(diagram_kwargs=diagram_kwargs)
-    _to_diagram_element(element, lookup=lookup, parent=None, vertical=vertical)
+    show_results_names = diagram_kwargs.pop('show_results_names', False) if diagram_kwargs is not None else False
```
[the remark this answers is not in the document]
ptmcg: Cleaning up - changes rippled up into ParserElement.create_diagram also.

pyparsing/diagram/__init__.py
multimeric: In that case I'm happy for you to change the name of the `first` attribute as well. I can't actually remember why I called it that, possibly because it stores the "first" copy of each element we encounter. But then it sounds like elements are copied when reused so this doesn't really make sense anyway.

ptmcg:
Great code comments, by the way, thanks for taking a close look!

pyparsing/diagram/__init__.py
ptmcg: I think there are two possible reasons why we might have a name annotation: if there is a defined results name on the current element, or if there was a name hint passed down from the parent. So we *might* see two annotations for the same element. I think this would do it:

    outer = Forward().setName("outer")
    inner = Word(alphas)[...].setName("inner")
    outer <<= inner

Rare, but not impossible.

pyparsing/diagram/__init__.py
ptmcg: Added

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
ptmcg: I just removed the hard-coded "base_expr" name - the diagram correctly shows the element as part of its containing parent.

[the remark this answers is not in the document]
multimeric: huh? How does that relate to this deduplication problem?

pyparsing/diagram/__init__.py
multimeric: Right, but in that scenario the name hint should get overridden no? Hence it just being a "hint".

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
ptmcg: It is a variation on the dedup problem. Since we are deduping on element name, we are relying on the developer not to do something like this:

    integer = Word(nums).setName("integer")
    signed_integer = Regex(r"[+-]\d+").setName("integer")  # should be a different name

Which I don't think is too big a thing to ask. But in infixNotation, my pyparsing code makes the calls to setName(), so it is not under the control of the developer. There are several calls to setName there, composing the name from the terms used in the infixNotation operators - there should be no collisions with these. But when I added a hard-coded "base_expr" name also, then this could create a duplicate if you had 2 or more infixNotations in the same parser, and the user wouldn't be able to fix it. (I might need to come up with a new pyparsing warning if a setName is called with the same name for two different elements.)

I added the "base_expr" when we hadn't fully addressed the Forward containment problem. Once that was resolved, I left it in, bu then realized I didn't need it anymore.

Sorry for being so long-winded. This is another case where there is a lot going on.

pyparsing/diagram/__init__.py
ptmcg: Okay, I just commented out that second block, and generated the diagram for that snippet above, and it still generates the desired diagram. So I'll delete it.

pyparsing/diagram/__init__.py
ptmcg: Are you suggesting a change here?

ptmcg:
I needed to add more test cases, and so I refactored the test_diagram.py module to move all the boilerplate code up to a common method. This also made it easier to drop in custom parsers in addition to pulling in examples.

I also worked out what I needed to do to handle Each elements - I ended up compositing a DiagramItem from a Group of a OneOrMore of a Choice (since AlternatingSequence is limited to just 2 choices).

So that is the end of my churn for today. I don't think there is much more I want to do here.

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
multimeric: Okay, I see now. But it still seems to me that this dictionary should be just indexed by element names if that is ultimately what we are using to dedup.

.travis.yml
```diff
@@ -17,6 +17,8 @@ matrix:
   fast_finish: true
 
 install:
+  - pip install railroad-diagrams
```
multimeric, line 20: These two dependencies are supposed to be installed as "extras" ie `pip install pyparsing[diagrams]`. If you can work out how to tell Travis about extras, perhaps using a custom `install` key, this would be more robust

pyparsing/core.py
```diff
@@ -2036,7 +2042,14 @@ def create_diagram(
                 "and jinja2 from https://pypi.org/project/jinja2 to generate parser railroad diagrams"
```
multimeric, line 2042: As with the Travis, I would prefer that we recommend that the users install the extra instead of the packages separately, as written in the docs: https://pyparsing-docs.readthedocs.io/en/latest/HowToUsePyparsing.html#id20

pyparsing/diagram/__init__.py
```diff
@@ -29,6 +30,32 @@
 T = TypeVar("T")
 
 
+class EachItem(railroad.Group):
+    """
+    Custom railroad item to compose a:
+    - Group containing a
+      - OneOrMore containing a
+        - Choice of the elements in the Each
+    with the group label indicating that all must be matched
```
multimeric, line 39: Didn't you decide [here](https://github.com/pyparsing/pyparsing/issues/226#issuecomment-913092001) that the Group labels are only used for result names?

pyparsing/diagram/__init__.py
```diff
@@ -102,26 +133,52 @@ def resolve_partial(partial: "EditablePartial[T]") -> T:
 
 def to_railroad(
     element: pyparsing.ParserElement,
-    diagram_kwargs: dict = {},
-    vertical: int = None,
+    diagram_kwargs: Optional[Dict] = None,
```
multimeric, line 136: I could be wrong, but I think when the `Dict` doesn't have defined types, you should use `dict` instead.

pyparsing/diagram/__init__.py
```diff
@@ -161,7 +218,7 @@ def __init__(
         #: The order in which we found this element, used for sorting diagrams if this is extracted into a diagram
         self.number = number  # type: int
         #: The index of this inside its parent
-        self.parent_index = index  # type: Optional[int]
+        self.parent_index = parent_index  # type: Optional[int]
```
multimeric, line 221: I guess since you've moved to Python 3.6 with type annotations, these can be pulled out as full annotations instead of type comments?

pyparsing/diagram/__init__.py
```diff
@@ -200,17 +259,30 @@ class ConverterState:
     Stores some state that persists between recursions into the element tree
     """
 
-    def __init__(self, diagram_kwargs: dict = {}):
-        #: A dictionary mapping ParserElement IDs to state relating to them
-        self.first = {}  # type:  Dict[int, ElementState]
+    def __init__(self, diagram_kwargs: Optional[Dict] = None):
```
multimeric, line 262: As above: `dict`

pyparsing/diagram/__init__.py
```diff
@@ -172,6 +229,9 @@ def mark_for_extraction(
     ):
         """
         Called when this instance has been seen twice, and thus should eventually be extracted into a sub-diagram
+        :param el_id:
```
multimeric, line 232: I guess in theory these should have descriptions

pyparsing/diagram/__init__.py
ptmcg: They are not quite exactly the same. One tests for the element_results_name and one tests for the group_results_name, potentially creating group annotations for both.

multimeric: I can see there's a snippet for that as well, but I believe these two are identical?

https://github.com/pyparsing/pyparsing/pull/298/files/97fcd3dbf6947d9df445357a499c99d5ede7e5a1#diff-bb7411dbf87d8b09728b60b573752816c42b0b4fc34f802d8209f75814a94b2bR381-R384

https://github.com/pyparsing/pyparsing/pull/298/files/97fcd3dbf6947d9df445357a499c99d5ede7e5a1#diff-bb7411dbf87d8b09728b60b573752816c42b0b4fc34f802d8209f75814a94b2bR481-R484

pyparsing/diagram/__init__.py
ptmcg: Actually, there was even a qualifier in that note about `NotAny`. Short of defining more extensive constructs, these annotations seemed to be the simplest way to add an indication of special matching behavior. To try to distinguish them from results names, I'm making the annotations all uppercase, and enclosed in `[]`'s. The exception cases are:
- `NotAny` - annotated with `[NOT]`
- `FollowedBy` - annotated with `[LOOKAHEAD]`
- `PrecededBy` - annotated with `[LOOKBEHIND]`

I was hoping that the brackets and uppercase might help distinguish these annotations from results names. Fortunately, it should be rare for any of these constructs to also have a results name.

multimeric: Okay, as long as everything is consistent I'm sure it will be fine.

pyparsing/diagram/__init__.py
ptmcg: Ah, okay.

Will do.

pyparsing/core.py
ptmcg: Yes! Will do.

pyparsing/diagram/__init__.py
ptmcg: In Py3.9, you don't even need Dict, you can just do `dict[str, Any]` or whatever. I was confusing that feature (which we can't use yet, for 3.6 compat) with your suggestion.

Will do.

pyparsing/diagram/__init__.py
ptmcg: Will do.

pyparsing/diagram/__init__.py
ptmcg: Picky, picky. :) 

Will do.

pyparsing/diagram/__init__.py
ptmcg: They *are* identical. Code from line 481 has to be copied up to the earlier bits because of the early 'return' statements.

pyparsing/diagram/__init__.py
multimeric: Yeah sorry this is a bit unnecessary. Normally for internal methods I just remove the parameter documentation entirely, I just don't like the look of a list of undocumented parameters I guess. But really it's up to you.

pyparsing/diagram/__init__.py
multimeric: Oh I see. Yes this redundancy bothers me a little. I wonder if that restructure wouldn't be too hard?

pyparsing/diagram/__init__.py
multimeric: Sorry I think I was wrong here. Neither `mypy` nor PyCharm shows a preference for `dict` over `Dict` in any case. I think as you say that in 3.9+ `dict` will be preferred but at this point it doesn't matter.

pyparsing/diagram/__init__.py
ptmcg: No, perfectly valid. If I leave them out, PyCharm highlights the parameters saying "you forgot to document these in the docstring". So easier to add the description (as long as they are accurate).

.travis.yml
ptmcg: For now, feels like more trouble than it is worth, since we manage both the travis.yml and the dependencies for diagrams.

multimeric: I think it should really be as simple as adding `install: pip install pyparsing[diagrams]`. The issue with this approach is that if these dependencies ever change, they would need to be updated in two different places.

multimeric: Or I guess since this stuff is done using tox (?), then using this  https://tox.readthedocs.io/en/latest/config.html#conf-extras

pyparsing/diagram/__init__.py
ptmcg: I just looked at it again, there is at least one more more early return further down.

We *could* wrap it all in a try-finally, so that post return, this code always be done. But that is kind of an abuse of try-finally.

I'll try a decorator - it doesn't look so bad, and allows us to delete the duplicated code.

.travis.yml
ptmcg: Oh, it's already there. So I'll just remove the lines from .travis.yml.

pyparsing/diagram/__init__.py
ptmcg: Decorator looks good. I also just tried an experiment where we would show if a parse action is present - not an annotation, but another element in the sequence. I'm not adding it now, it was just to see how this would work, and this decorator is a nice place to hook in other types of such "embellishments".

Here's how it looks adding the parse action indicators (a block with a bold asterisk). It helps show why infixNotation creates the extra paths the way it does.
Web capture_7-9-2021_104239_

Again, I'm not proposing this feature now. Just showing that we could use the decorator for other such items.

ptmcg: I'll commit the decorator, see what you think.

pyparsing/diagram/__init__.py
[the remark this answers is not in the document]
ptmcg: Not really a conflict here. We need to keep separate entries by element id in the lookup to keep straight which element has which results name, so they get displayed properly in the parent element's diagram. When we are deduping at the end, we are deduping by the subdiagram _name_ - there is no reason to repeat the same subdiagram just because it occurred in two separate places in two separate elements with the same name and description.

pyparsing/diagram/__init__.py
ptmcg: Ha! Once I moved the code to the decorator, the group_results_name turned out to be unused, so I just cut it. Less confusing cruft = better code.

ptmcg:
I think the code is in releaseable shape now (plus this PR has gotten pretty long). I felt like this was a good way to collaborate across time and space. Thanks for all the comments.

pyparsing/diagram/__init__.py
```diff
@@ -102,26 +133,55 @@ def resolve_partial(partial: "EditablePartial[T]") -> T:
...
     # Now that we're finished, we can convert from intermediate structures into Railroad elements
-    resolved = [resolve_partial(partial) for partial in lookup.diagrams.values()]
+    diags = list(lookup.diagrams.values())
+    if len(diags) > 1:
+        # collapse out duplicate diags with the same name
+        seen = set()
+        deduped_diags = []
+        for d in diags:
+            # don't extract SkipTo elements, they are uninformative as subdiagrams
+            if d.name == "...":
```
multimeric, line 175: If you want these to never be extracted, consider adding this functionality to the `worth_extracting` function, which encapsulates just this: https://github.com/pyparsing/pyparsing/blob/154d3d14563a3145ba13583d29cab6dff3e9c037/pyparsing/diagram/__init__.py#L338.

multimeric (approved):

multimeric:
Yeah looks good. I could nitpick on this endlessly but really if it does what you want and you understand it then that's for the best.

PR #334, opened 2021-11-10
Add equality comparison to ParseResults

ThoFrank:
Hi there,
I really enjoy using this library! Thanks for your great work

While implementing a remembering ParserElement I missed the ability to check two ParseResults for Equality. So I added it.

### DIFF
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -250,6 +250,12 @@ class ParseResults:
     def __reversed__(self) -> Iterator:
         return iter(self._toklist[::-1])
 
+    def __eq__(self, o: object) -> bool:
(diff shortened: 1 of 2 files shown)


ThoFrank:
Ah I guess that won't work as ParseResults is mutable and therefor cannot be hashable...

ThoFrank:
Thanks for looking into it.
Yeah I totally understand that adding the __eq__ method doesn't make much sense for the project. Also I realized that a ParserElement that remembers what it encountered previously doesn't make too much sense because of the false positives it can produce, when it remembers a element parsed by a failing grammar rule.
So I don't really have a need for that functionality any more...

Issue #330, opened 2021-11-08
No limits on ZeroOrMore/OneOrMore?


elespike:
Hello,

Is there a way to limit the maximum number of matches when using `ZeroOrMore`, `OneOrMore`, or `delimited_list`?

Otherwise, parsers can easily be vulnerable to resource consumption:
```pycon
>>> from pyparsing import delimited_list
>>> d = delimited_list('A')
>>> x = d.parse_string('A,' * 1000000000000000000)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
MemoryError
```

I'm aware of the `[x, y]` construct, but not sure if I could modify an existing parser expression (like the return value from `delimited_list`) to add those limits.

elespike:
🤦‍♂️ good catch lol
I did see the script hanging with lower numbers, but thought the MemoryError was more visible for an example 😅 didn't consider it wasn't related!

Thanks for looking into it, glad it helped surface some bugs. =)

What about limiting the size of a `delimited_list`? Will `[x, y]` work with its returned expression?

ptmcg:
No, you need to limit the work during parsing, not after the fact. You can usually step through the results from a large parse using scan_string, since it returns a generator. But within the grammar, repeaters like `ZeroOrMore`, `OneOrMore`, and `delimitedList` just keep going as long as their underlying expression keeps matching the input.

You could try implementing a custom version of `OneOrMore` that keeps a count of the matches, and then stops at some limit, or raises an exception when the limit is reached, or discards any more matches after reaching the limit (i.e., keep parsing, but don't blow up memory by adding any more matches to the parsed results).

What value of `y` were you thinking of? If `y > sys.recursionlimit/4` and using `[..., y]` , you'll get the `RecursionError`.

elespike:
🤔 I see, makes sense to limit with `scan_string` and that's actually what I happen to be using. I'll go that route.

For my purposes, a `y` of 99 would be more than enough. My initial idea was to implement a clone of `delimited_list` that used `[0, 99]` instead of `ZeroOrMore`. I got really close, but for some reason I couldn't get `set_results_name` to work with my implementation. 🤷 but I don't want to go that route anyway.

Then I thought maybe I'd submit a patch for an added kwarg for `delimited_list`. After this discussion, doesn't sound like that's the way to go. 😆 

I should be able to handle it with the generator 👍 thanks for your help!

elespike:
Hm... don't think it'll work. =/

```pycon
[...]

PR #335, opened 2021-11-11
Added min/max keyword arguments for delimited_list

elespike:
As discussed in #330, here's a patch to add `max` to `delimited_list`.

A few notes/questions:
- Added `min` in addition to `max`, let me know if you'd rather only have `max`
- Used the names `_min` and `_max` to avoid collision with Python built-ins
- Currently, `delimitedList('A', _max=0)` has the same effect as `delimitedList('A', _max=1)`. Same for `_min=0`/`_min=1`.
  Should `_min=0`/`_max=0` raise a `ValueError` instead?

### DIFF
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -14,6 +14,8 @@ def delimited_list(
     expr: Union[str, ParserElement],
     delim: Union[str, ParserElement] = ",",
     combine: bool = False,
+    _min: OptionalType[int] = None,
+    _max: OptionalType[int] = None,
     *,
     allow_trailing_delim: bool = False,
 ) -> ParserElement:
@@ -46,7 +48,12 @@ def delimited_list(
     if not combine:
         delim = Suppress(delim)
 
-    delimited_list_expr = expr + ZeroOrMore(delim + expr)
+    if _min is None and _max is None:
+        delimited_list_expr = expr + ZeroOrMore(delim + expr)
+    else:
+        _min = ... if _min is None else max(0, _min - 1)
+        _max = ... if _max is None else max(0, _max - 1)
+        delimited_list_expr = expr + (delim + expr)[_min, _max]
 
     if allow_trailing_delim:
         delimited_list_expr += Opt(delim)
--- a/pyparsing/helpers.py
+++ b/pyparsing/helpers.py
@@ -14,8 +14,8 @@ def delimited_list(
     expr: Union[str, ParserElement],
     delim: Union[str, ParserElement] = ",",
     combine: bool = False,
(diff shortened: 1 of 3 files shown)


pyparsing/helpers.py
```diff
@@ -46,7 +48,12 @@ def delimited_list(
     if not combine:
         delim = Suppress(delim)
 
-    delimited_list_expr = expr + ZeroOrMore(delim + expr)
+    if _min is None and _max is None:
+        delimited_list_expr = expr + ZeroOrMore(delim + expr)
+    else:
```
ptmcg, line 53: Instead of this code, let's just value check the values, something like this:

    if None not in (min, max) and not (0 < min <= max):
        raise ValueError("min and max values must  be greater than 0, and min <= max")
    else:
        # one or both values are None, so no need to compare relative to each other
        if min is not None and min < 1:
            raise ValueError("min must be greater than 0")
        if max is not None and max < 1:
            raise ValueError("max must be greater than 0") 

Will probably want to add tests in unit_tests.py to make sure we get expected exceptions.

The value check now guards against getting min or max equal to 0, so can probably simplify to:

    _min = ... if min is None else min-1
    _max = ... if max is None else max-1

ptmcg (changes requested):

ptmcg:
I can go either way on adding `min` as well as `max`, but as you have found, it is much less complicated to just add `max`. :)

In general, pyparsing doesn't really ever see `min=0` type expressions. `Word("a", min=0)` is more properly written as `Optional(Word("a"))`, and if omitted `min=1` is just assumed. I should go back and check the `Word` code to see if these `ValueError` checks might simplify some of that code too. (`Word` also has `exact`, which makes things even more complicated - let's not add `exact` to `delimited_list`.)

ptmcg:
Thanks for taking the time to work this up, btw!

elespike:
My pleasure, and thanks for the feedback - will get on it =)

elespike:
Quick question: is it necessary to keep the `ZeroOrMore` case? If we make `min=1, max=None` the defaults, that's the same as using `ZeroOrMore` with a leading `expr`, and can be written as `expr + (delim + expr)[min, max]` (assuming that recursion limit won't interfere).

ptmcg:
Today got sucked up into family tasks, so tomorrow is Catch-up Day - will get back to looking at this over the weekend.

elespike:
:+1: no rush =)

elespike:
@ptmcg just checking in, in case you've had a chance to see the latest changes

ptmcg:
Overall looks good - thanks for the ping!

PR #342, opened 2021-12-06
Fixing Unicode block range in examples/booleansearchparser.py

tc-yu:
Also updated a few methods to use the snake case version


tc-yu:
fixes # 339

PR #347, opened 2021-12-21
update bigquery SQL parser

fadimaali:


fadimaali:
sorry didn't realise this is created against the main repo. Will clean up then open again

examples/bigquery_view_parser.py
```diff
@@ -67,106 +67,112 @@ def _get_parser(cls):
...
-        )
-        # fmt: on
+        # those are keywords that are fiunction names
+        keyword_funcs = MatchFirst((
+            TREAT, WITHIN, TABLESAMPLE, SOME, SET, RIGHT, PROTO, MERGE,
+            LOOKUP, LEFT, LATERAL, FETCH, CURRENT
+        ))
 
-        identifier_word = Word(alphas + "_@#", alphanums + "@$#_")
+        identifier_word = Word(alphas + '_@#', alphanums + '@$#_')
```
ptmcg, line 137: Please don't undo black formatting.

examples/bigquery_view_parser.py
```diff
@@ -67,106 +67,112 @@ def _get_parser(cls):
...
         collation_name = identifier.copy()
-        # NOTE: Column names can be keywords.  Doc says they cannot, but in practice it seems to work.
-        column_name = identifier_word.copy()
-        qualified_column_name = Combine(
-            column_name + ("." + column_name)[..., 6], adjacent=False
-        )
+        # NOTE: Column names can't be keywords unless they are quoted
+        column_name = identifier.copy() | Suppress('`') + identifier_word + Suppress('`')
+        # first part of multi part column name can't be keyword, other parts can
+        qualified_column_name = Combine(column_name + (
```
ptmcg, line 143: Instead of adding `ZeroOrMore(' ')`, can you do just do `Combine` with `adjacent=False`? (What if there is a line break instead of a space, for instance?)

examples/bigquery_view_parser.py
```diff
@@ -67,106 +67,112 @@ def _get_parser(cls):
...
+        # SQL like `SELECT a, FROM b` becomes ambiguous if we support `from`.  In that SQL, is `FROM` a column name with
+        # alias `b`, or are we selecting a single column from table `b`?
+        column_alias = identifier.copy()
         table_name = identifier.copy()
         table_alias = identifier.copy()
         index_name = identifier.copy()
-        function_name = identifier.copy()
+
+        standard_name_part = ~keyword + Word(alphanums + "_" + "-") | keyword_funcs
+        quoted_name_part = Suppress("`") + CharsNotIn("`") + Suppress("`")
```
ptmcg, line 157: Probably good to define a QuotedString("\`") expression and reuse it in place of `Suppress+CharsNotIn+Suppress`

ptmcg:
Purpose of sorting the keyword lists?  Makes it difficult to add new keywords later (instead of just appending, may require insertion, with resultant changes to following lines of unchanged code).

Please do the following:
- run black on your code before committing
- don't remove trailing commas in lists

These are a lot of changes, some just reformatting, some of actual changes to the grammar - maybe open an issue first describing the non-reformatting changes you propose to make.

fadimaali:
Thanks @ptmcg . I created an issue https://github.com/pyparsing/pyparsing/issues/348 and working on cleaning up the PR and opening it again

PR #349, opened 2021-12-23
Add test cases

fadimaali:


ptmcg:
Maintenance of this example has become more extensive than I care to invest in an example. At this point, I suggest you break it out to be its own project.

PR #364, opened 2022-02-14
Mark `pyparsing` as a typed package

BvB93:
Packages have to distribute a `py.typed` file in order to signal type checkers that the package is typed (see [PEP 561](https://www.python.org/dev/peps/pep-0561/#packaging-type-information)).
This file was previously missing though, meaning that downstream users were unable to benefit from pyparsings' type annotations.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,10 @@ setup(  # Distribution meta-data
     extras_require={
         "diagrams": ["railroad-diagrams", "jinja2"],
     },
-    package_data={"pyparsing.diagram": ["*.jinja2"]},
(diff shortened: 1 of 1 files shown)


eggplants:
When do pyparsing make new release including this commit?
Currently mypy raises this errors in all typed packages using pyparsing:

```
rdflib/plugins/sparql/evaluate.py:24: error: Skipping analyzing "pyparsing": module is installed, but missing library stubs
or py.typed marker  [import]
    from pyparsing import ParseException
```

I can't wait for publishing pyparsing as a typed package!

PR #371, opened 2022-03-16
Add missing type hints

heckad:


aucampia:
Happy to see some type hints, but I don't see any type checker as part of the CI pipeline, should this not be in place first?

aucampia:
Let me raise a separate issue for that actually.

heckad:
I added only those types that help pass typechaking in my project. I am sure that after a new version with a `py.typed` file is released, the community will quickly add the missing types. And only then it will be possible to think about checking the project itself and arranging internal types. Everything big starts from small.

ptmcg:
Sorry for the delay - was on a business trip, so was focused on the day job.

I also appreciate your comments on when/how pyparsing might _someday_ get type checking into the CI, but it won't be for a while yet. (Since I still support Py3.6, I can't even add `from __future__ import annotations`)

heckad:
@ptmcg, do you have any plans for a new release?

PR #402, opened 2022-05-28
Fix type annotations of Forward dunder-methods

sirosen:
The `__lshift__`, `__ilshift__`, and `__or__` methods each return a ParserElement object, but have no annotated return type. The result is that the following code will not type check:

    def foo() -> pp.ParserElement:
        expr = pp.Forward()
        expr <<= bar()
        return expr | pp.Literal("baz")

whereas the code will type check if the return line is changed to

    return pp.MatchFirst([expr, pp.Literal("baz")])

This is a bug in the types which can be resolved fairly simply with some return type annotations.

Testing is more complicated. Testing annotation accuracy is a relatively novel space with a few options, none of which can be considered standard as of yet.

Many solutions require learning a new toolchain only for that purpose. However, one of the lower-impact options is to use `mypy --warn-unused-ignores` to check that annotations satisfy some constraints. This isn't the most precise test possible, but it's simple and uses a widely known and familiar tool for the job.
`tox -e mypy-tests` is a new tox env which calls `mypy` in the desired way. We can confirm with a new test case file that `tox -e mypy-tests` fails prior to this change to `pyparsing/` and that it passes with the change made.

---

Some follow-up notes.
- The testing strategy is documented [in this section of the typing readthedocs](https://typing.readthedocs.io/en/latest/source/quality.html), but I contributed that guide to the docs, so there's authorship bias.
- `mypy` doesn't run cleanly from the repo root. I believe it's importing the local `pyparsing` package tree and failing on it. If this is a concern, moving to a `src/` tree setup or making `mypy` pass on `pyparsing/` would work, but I don't think anything simpler would suffice. (`--exclude` does not work.)
- I put the mypy test files in `tests/` but could use a different dir if there's any concern about file naming and `unittest` discovery. Under pytest, I've used norecursedirs -- I don't think unittest has a similar option.
- The pin in `tox` to `mypy==0.960` is intentional, as I've seen issues in other projects when mypy does a release and builds start failing due to a new check. This could be done in a requirements.txt if that's desired, of course.
- I think it would be a good idea to work up a `mypy.ini` config which makes `mypy pyparsing/` pass. Any typing changes can be gradual -- I've seen this mentioned in another thread here -- but publishing types without validation passing, even with some permissive options, leaves the door open to weird things. I'd be happy to help with this and the minimal additional annotations it would probably require.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -5178,7 +5178,7 @@ class Forward(ParseElementEnhance):
         super().__init__(other, savelist=False)
         self.lshift_line = None
 
-    def __lshift__(self, other):
+    def __lshift__(self, other) -> "Forward":
         if hasattr(self, "caller_frame"):
             del self.caller_frame
         if isinstance(other, str_type):
@@ -5195,10 +5195,10 @@ class Forward(ParseElementEnhance):
         self.lshift_line = traceback.extract_stack(limit=2)[-2]
         return self
 
-    def __ilshift__(self, other):
+    def __ilshift__(self, other) -> "Forward":
         return self << other
 
-    def __or__(self, other):
+    def __or__(self, other) -> "ParserElement":
         caller_line = traceback.extract_stack(limit=2)[-2]
         if (
             __diag__.warn_on_match_first_with_lshift_operator
--- a/tests/README.md
+++ b/tests/README.md
@@ -8,3 +8,12 @@ After forking the pyparsing repo, and cloning your fork locally, install the lib
 Run the tests to ensure your environment is setup 
      
      python -m unittest discover tests
+
+### mypy ignore tests
+
+`tests/mypy-ignore-cases/` is populated with python files which are meant to be
+checked using `mypy --warn-unused-ignores`.
+
+To check these files, run
+
+    tox -e mypy-tests
--- a/tests/mypy-ignore-cases/forward_methods.py
+++ b/tests/mypy-ignore-cases/forward_methods.py
@@ -1,0 +1,14 @@
+import pyparsing as pp
+
+# first, some basic validation: forward is a ParserElement, so is Literal
+# MatchFirst([Forward(), Literal(...)]) should also be okay
+e: pp.ParserElement = pp.Forward()
+e = pp.Literal()
+e = pp.MatchFirst([pp.Forward(), pp.Literal("hi there")])
+# confirm that it isn't returning Any because it cannot be assigned to a str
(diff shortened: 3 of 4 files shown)


ptmcg:
Thanks for looking at this further.  You may be underestimating the amount of "additional annotations it would probably require."  :)

sirosen:
Haha, I may be underestimating it, even expecting it to be large! I am happy to try though, to see what I can do to help out.

I'm looking at the CI failure for pypy and I don't really understand what it means. I'll try to make time later today to run tests on pypy locally to see if I can reproduce. It seems unlikely to me that this PR would break any runtime tests, however.

ptmcg:
It looks to me like it was some kind of timeout, more than an actual test error. See if this helps: https://stackoverflow.com/questions/59073731/set-default-timeout-on-github-action-pipeline

After you clear up the CI issues and this merges successfully, please let this rest for a day. I have a number of type annotations queued up in a Git shelf that I can revisit/add, and then you can refresh your forked repo and continue with the difficult ones.  Just trying to avoid our stepping on each other.

ptmcg:
And these _are_ a big help, by the way, much appreciated.

ptmcg:
Looks like mypy and PyPy aren't getting along. Let's not do the mypy checks with PyPy. I'm pretty good with this as long as the regular tests run with PyPy, we'll know we haven't broken anything at actual run time. And the type checks run on other versions will give us sufficient type feedback. (A thought: do we even need to run mypy on multiple versions of CPython?)

ptmcg:
Another option is for us to make these type annotation changes, but continue to run mypy locally and not as part of CI until work is further along.

sirosen:
I was working on this yesterday and I came to the same conclusion -- pypy doesn't seem to be able to install mypy.

I wanted to get a passing build, so I tooled around on my fork and produced this change to the workflow:
https://github.com/sirosen/pyparsing/commit/50e028d2e4d24054dcc5cc3081b17763cf32c66d

The main issue with the current workflow file is that `TOXENV` is being set in the builds, but then `tox -e ALL` is being used. If we use `TOXENV=py tox` for the builds, we get nicer behavior because tox only runs the main testenv under the current python. I turned this into a matrix config to ensure that `mypy-test` and `pypy` are never combined.

I _could_ push that as a commit on this branch, which would expand this PR much beyond its original intent to be a CI refactor. So what I'm going to do right now is comment out the `mypy-test` env in `tox.ini`. I can then submit CI changes to run like this in a new, separate PR.

PR #407, opened 2022-06-04
Fix list formatting in docstrings

djpohly:
A few docstrings were missing a blank line before lists, causing the start of the list to be parsed by RST as a continuation of the previous line.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -594,9 +594,9 @@ class ParserElement(ABC):
         Each parse action ``fn`` is a callable method with 0-3 arguments, called as
         ``fn(s, loc, toks)`` , ``fn(loc, toks)`` , ``fn(toks)`` , or just ``fn()`` , where:
 
-        - s   = the original string being parsed (see note below)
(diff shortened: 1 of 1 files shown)


ptmcg:
Nice catch thanks.

While looking at the changes, I see that several of the parameter lists omit the `` backticks on the parameter names (ParserElement.run_tests is correct, ParserElement.create_diagram is not - missing backticks). 

Since you are cleaning up the docstrings, would you mind going through and cleaning these up as well?

djpohly:
No problem! Other docstring inconsistencies I've noticed:

* The default value for parameters is sometimes given in parentheses before the description, sometimes after, sometimes noted inline, and sometimes not mentioned.
* Sometimes the parameter type is given in the description, sometimes not.
* Some functions use Sphinx param/raises/returns directives, others do not.

I can't tell what the preferred choice is for each of these, but I'd be happy to add some consistency while I'm updating the documentation.

ptmcg:
I'm happy merging what you've done so far, and picking up other docstring cleanup in another PR.

PR #410, opened 2022-06-09
fix Sphinx errors/warnings

djpohly:
Two warnings remain from the Hebrew and Devanagari names in pyparsing_unicode, but those are due to Sphinx using Python's builtin `re` library to parse identifiers (which does not have thorough Unicode handling for `\w`).


ptmcg:
Thanks!

Issue #412, opened 2022-06-12
Support `... | ""` as alias for `Opt(...)`


anntzer:
Depending on the context, `<something> | ""` (or `"" | <something>`) may be a clearer way to spell out `Opt(...)`; I believe this *almost* works except for the fact that `Literal("")` currently raises an error asking to use `Empty()` instead.  Could this restriction be lifted?

I see various ways to lift the restruction, depending on what you are exactly willing to accept.  One would be to just directly allow `Literal("")` (if you worry about the loss in perf, you *could* define `Literal.__new__` to return `Empty()` in that case, although that's a bit too magical in my taste).  Or you could change `ParseElementEnhance.__init__` at https://github.com/pyparsing/pyparsing/blob/966d6fded149c6c11993746b0d72166bc04e4504/pyparsing/core.py#L4345-L4353 to special-case `expr == ""` there (or rather `len(expr) == 0` to support bytes too) and return `Empty()`.  Finally, you could also change `ParserElement.__or__/__ror__` to detect that case and construct an `Optional(...)` as needed (this may be the simplest from an explanation point of view: "`|` constructs a MatchFirst, except that `| ""` constructs an Optional").

I can make a PR if you agree with any of these solutions.

ptmcg:
I'll look these ideas over, but it won't be for a few days.

Could you put together some examples showing how `expr | ""` or `expr | Empty()` is preferable to `Opt(expr)`?  (Granted, this seems to be paralleling the discussion around `typing.Optional[SomeType]` being better written as `SomeType | None`.) 

Doesn't `expr | Empty()` work with the current code, without making any changes to Literal?

I guess `Empty()` is not quite so easy to write as `""`, and requires an import, or writing as `pp.Empty()`. What about `expr | None`? And have the `__or__` operator recognize a `None` operand and replace it with `Empty()`?  (Still need to give this a little more thought, but this seems like it might address your use case.)

anntzer:
For example, consider matplotlib's tex-parser element for tex's genfrac command (`\genfrac{left_delim}{right_delim}{rulesize}{style}{num}{denum}`) -- not a full example, other `p.foo` elements are defined earlier:
```python
        p.genfrac <<= r"\genfrac" - (
            "{" + Optional(p.ambi_delim | p.left_delim)("ldelim") + "}"
            + "{" + Optional(p.ambi_delim | p.right_delim)("rdelim") + "}"
            + "{" + p.float_literal("rulesize") + "}"
[...]

PR #413, opened 2022-06-16
Use Literal.__new__ to select optimized subclasses

djpohly:
This turns `Literal()` into a factory which creates an object of the appropriate type from the start, rather than having to overwrite the `__class__` attribute later.

See #412 for the suggestion.

Question: should `Empty` be made a subclass of `Literal`?  It seems to make sense that it would be.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -2343,7 +2343,10 @@ class Literal(Token):
     def __copy__(self) -> "Literal":
         # Needed to assist copy.copy() (used in ParserElement.copy), which
         # doesn't handle the factory __new__ well.
-        return Literal(self.match)
+        obj = Literal(self.match)
+        # Copy instance attributes
+        obj.__dict__.update(self.__dict__)
+        return obj
 
     def _generateDefaultName(self) -> str:
         return repr(self.match)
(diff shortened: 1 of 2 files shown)


ptmcg:
> Question: should Empty be made a subclass of Literal? It seems to make sense that it would be.

If `Literal.__new__` can return an Empty, then that seems logical that Empty would be a subclass of Literal. Go ahead and include that change too (it should also help some of the internal type checking).

pyparsing/core.py
```diff
@@ -2331,11 +2319,9 @@ def __new__(cls, match_string: str = "", *, matchString: str = ""):
     def __init__(self, match_string: str = "", *, matchString: str = ""):
         super().__init__()
         match_string = matchString or match_string
-        if not match_string:
-            raise ValueError("Empty string should not reach Literal.__init__()")
         self.match = match_string
         self.matchLen = len(match_string)
-        self.firstMatchChar = match_string[0]
+        self.firstMatchChar = match_string[0] if match_string else ''
```
ptmcg, line 2324: Could also write as `self.firstMatchChar = match_string[:1]`

ptmcg: Also, run `black`, which will change all string literals to using "" instead of ''

ptmcg:
Also nice to get rid of the assignment to `__class__`, which always gets me some dirty looks.

PR #418, opened 2022-06-21
Fix a potential wrong operator bug

LostBenjamin:
Hi,

This pull request is a fix to a potential wrong operator bug at `pyparsing/core.py`. Please check the changes.

Best,
Jingxuan

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -3965,7 +3965,7 @@ class And(ParseExpression):
                     )
             else:
                 loc, exprtokens = e._parse(instring, loc, doActions)
-            if exprtokens or exprtokens.haskeys():
+            if exprtokens and exprtokens.haskeys():
                 resultlist += exprtokens
         return loc, resultlist
 


ptmcg:
This is a really interesting PR, though not for the change you have suggested. In version 3.0, the determination of `bool(results)` where results is a `ParseResults` object, changed from version 2.x.  This changed in the `ParseResults` class, but related code did not get updated.

There are actually 2 instances of this code, one at line 3968, which you found and corrected, and another at line 4889.  These can now be changed to simply `if exprtokens:` and `if tmpTokens:`, because the new v3.0 `__bool__` code does the check for the presence of keys.

_BUT_

I think it would be better if we moved this logic into the `ParseResults.__iadd__` function. If you insert into the top of that function this code:

```
if not other:
    return self
```

Then the 2 cases can be simplified. Neither one has to test for whether exprtokens or tmpTokens is non-empty, and can just do the += augmented assignment in the if-body.

What do you think?

ptmcg:
Ok - yes this was helpful! I can definitely see how a code scanner would complain about that code! (What inspection tool are you using? I would mind adding it to my own code scanning process.)

I'll close this PR as Wont Fix, but the follow up work I described will show up in the next release.

LostBenjamin:
We use our latest research published in ICML'22. We have released our [paper](https://arxiv.org/abs/2204.10049) and [code](https://github.com/eth-sri/learning-real-bug-detector).

Issue #416, opened 2022-06-18
Parser element custom name isn't appropriate for `__repr__`


djpohly:
The Python [documentation on `__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__) encourages developers to return something that "look[s] like a valid Python expression that could be used to recreate an object with the same value" or that is at least "information-rich and unambiguous".

The custom name from `set_name` isn't really appropriate for this.  Always using the shorthand "default name" for `__repr__` would be more information-rich, although something that works like this would really be ideal:
```
>>> p = ZeroOrMore(one_of(["hello", "world"]) + "foo", stop_on=Keyword('end'))
>>> print(p)  # calls str()
[{hello | world 'foo'}]...
>>> p  # calls repr()
(Regex('hello|world') + 'foo')[...: Keyword('end')]
>>> q = eval(repr(p))  # nice!
>>> q.set_name("custom")  # name doesn't affect repr()
(Regex('hello|world') + 'foo')[...: Keyword('end')]
>>> print(q)  # name does affect str()
custom
```
All of the above I have working so far for a handful of ParserElement types.  Interested in a complete PR?

(And since I seem to tie everything into documentation: Sphinx uses `repr()` to show the values of documented variables.)

ptmcg:
This looks pretty cool! I think I looked at doing something like this a few years ago but, it got back-burnered in favor of other things.

Would the repr also include some of the other qualifiers like "set_name" or "set_parse_action"? Or just reconstruct the bare expression?

Yes, a PR to implement this would be very welcome.

djpohly:
I wasn't planning to show too much more than the required constructor arguments, lest it become unreadable too quickly. (The `eval` line above was meant only as a cute illustration, not as a proposed alternative to a class-aware `copy()` method.)

I'll polish something up for a PR.

djpohly:
PR is up, with a couple of questions:

1. `_PendingSkip` makes perfect sense in an `And` context, but I don't understand what it means to create one with the match-first `|` operator.  (I wasn't able to find it in the documentation or figure it out from the unit tests.)  What is `X | Y | ...` supposed to mean?
[...]

PR #423, opened 2022-07-01
Meaningful repr for ParserElements

djpohly:
See #416 for discussion.

Populated instances of `Forward` are represented only as `"Forward(...)"` to prevent infinite recursion.  I don't imagine this changing.

Helper classes like `_PendingSkip` and `_ErrorStop` are not yet handled, and there is no special handling for `Literal`.  I'll ask about these in the issue.

I held off on creating unit tests for the new implementations to let the devs decide whether `__repr__()` actually belongs in the test suite or not.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -476,7 +476,50 @@ class ParserElement(ABC):
         self.callDuringTry = False
         self.suppress_warnings_: List[Diagnostics] = []
 
-    def __repr__(self):
+    def __repr__(self) -> str:
+        """Returns a canonical string representation of this element.  This is
+        often, but not always, an expression that would produce a similar
+        element if evaluated.
+
+        Instances of the literal wrapper class (set by
+        :meth:`inline_literals_using`) are written as literals when possible.
+        """
+        return self._make_repr()
+
+    def __format__(self, spec: str) -> str:
+        """Formats an element for printing or f-string interpolation.
+
+        Supported conversion specifiers are:
+
+        - ``s`` or none: as given by :meth:`__str__`
+        - ``r``: as given by :meth:`__repr__`
+        - Other specifiers are used internally and may change.
+
+        Example::
+
+            parser = OneOrMore("Hello") + "world"
+            print(f"{parser}\\n{parser:r})
+
+        results in::
+
+            {{'Hello'}... 'world'}
+            Literal('Hello')[1, ...] + 'world'
+        """
+        if not spec or spec == "s":
+            return str(self)
+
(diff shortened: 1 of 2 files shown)


pyparsing/core.py
```diff
@@ -2299,7 +2301,7 @@ def must_skip(t):
             def show_skip(t):
                 if t._skipped.as_list()[-1:] == [""]:
                     t.pop("_skipped")
-                    t["_skipped"] = "missing <" + repr(self.anchor) + ">"
+                    t["_skipped"] = f"missing <{self.anchor}>"
```
ptmcg, line 2304: Used to show `repr(self.anchor)`  - should change to `{self.anchor!r}`"?

ptmcg:
Does this make the railroad diagrams better? worse?

ptmcg:
If you are using features (such as typing annotations) not yet available until Py 3.7, I'll be dropping Py3.6 support in pyparsing 3.1, but this won't be for a few months yet.

pyparsing/core.py
```diff
@@ -1408,7 +1412,8 @@ def __sub__(self, other) -> "ParserElement":
             raise TypeError(
                 f"Cannot combine element of type {type(other).__name__} with ParserElement"
             )
-        return self + And._ErrorStop() + other
+        # Construct And directly to ensure that an _ErrorStop is never the last element
```
ptmcg, line 1415: Interesting...

pyparsing/core.py
djpohly: This change was so that an [existing unit test](https://github.com/pyparsing/pyparsing/blob/81605f7ddc3a6217fffc57a71a93c72cc83e2d41/tests/test_unit.py#L1620) would pass. I'm still hazy on the what/why of `SkipTo` in this context - should the test be changed instead?

pyparsing/core.py
djpohly: It simplified the code in `And.__repr__()` a little to note that, in the only place where an `_ErrorStop` is created, it has another element after it.  If the user is creating an instance of an underscore-prefixed class, be it on their own head. :)

I guess it would still be possible to have a dangling `_ErrorStop` if `ParseExpression.exprs` is considered public API.  I can add a check for this if you'd like.

djpohly:
Looks like railroad diagrams must use `str()` - the ones generated by the tests, at least, were identical before and after the changes.

IIRC everything worked fine with `tox -e py36`.  Did I miss something?

ptmcg:
Let's close this PR, and start fresh from the current state of things if you are still interested.

Thanks a bunch for your contributions so far!!!

PR #425, opened 2022-07-04
Return NotImplemented for unsupported operations

### LINKED ISSUE #424: Operators should return `NotImplemented` rather than raise `TypeError`
The overloaded operators for `ParserElement`s should [return `NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented) rather than raising an exception, unless the intent is to prevent anyone from implementing other types which can be combined with `ParserElement`s.  (Even then, you can only stop it if you're the first operand.)  For a silly example:
```python
>>> class ThreeSir:
...   def __add__(self, other):
...     return 3
...   def __radd__(self, other):
...     return 3
...
>>> five = ThreeSir()
>>> "1 2" + five
3
>>> 1 + 2 + five
3
>>> Literal("three") + five
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/djpohly/src/pyparsing/pyparsing/core.py", line 1395, in __add__
    raise TypeError(
TypeError: Cannot combine element of type ThreeSir with ParserElement
>>> five + Literal("three")
3
```

djpohly:
Fixes #424.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1378,9 +1378,7 @@ class ParserElement(ABC):
         if isinstance(other, str_type):
             other = self._literalStringClass(other)
         if not isinstance(other, ParserElement):
-            raise TypeError(
-                f"Cannot combine element of type {type(other).__name__} with ParserElement"
-            )
+            return NotImplemented
         return And([self, other])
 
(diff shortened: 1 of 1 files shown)


ptmcg:
Excellent, thanks!

PR #427, opened 2022-07-07
Simpler fix for __new__ with copy.copy()

djpohly:
I discovered something more straightforward than what I had in #413 to fix the interactions between `copy.copy()` and a customized `__new__`.  (For info about the `__getnewargs__` method, see the [pickle module](https://docs.python.org/3/library/pickle.html#pickling-class-instances).)


ptmcg:
Ok, we'll give it a shot!

PR #437, opened 2022-08-12
Copy expression in ParseElementEnhance on ignoring

grv87:
To avoid the side effect that a pattern set to be ignored in one expression gets ignored everywhere else.

I didn't open an issue, since the problem and the fix seemed pretty trivial.

BTW, `leave_whitespace` and `ignore_whitespace` first make a copy, and then check for `None`. I don't think that's correct:

https://github.com/pyparsing/pyparsing/blob/d93930308f7fe79f2290812074f62a472c83db59/pyparsing/core.py#L4501-L4502

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -4517,11 +4517,11 @@ class ParseElementEnhance(ParserElement):
             if other not in self.ignoreExprs:
                 super().ignore(other)
                 if self.expr is not None:
-                    self.expr.ignore(self.ignoreExprs[-1])
+                    self.expr = self.expr.copy().ignore(self.ignoreExprs[-1])
         else:
             super().ignore(other)
             if self.expr is not None:
-                self.expr.ignore(self.ignoreExprs[-1])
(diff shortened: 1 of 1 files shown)


ptmcg:
I haven't jumped on this change because it looks like it is going to have a negative impact on performance, essentially defeating packrat parsing. Packrat relies heavily on detecting re-parsing the same expression at the same string location, and these copies would make it look like the same expression was different if it is in the ignore list.

I *will* look at that check for None after making a copy - you're right, that looks out of order.

ptmcg:
I did fix those checks for `None` in `leave_whitespace` and `ignore_whitespace`, thanks. (Released in 3.1.0a1)

As for making copies, I am very reluctant to make your proposed change, since it so adversely affects packratting. A workaround would be to make a copy of the element in the parser definition in the case where there would be different treatment of whitespace for the same expression.

grv87:
In short, it stays another undocumented inconsistency.

PR #446, opened 2022-10-28
Minor correction to copied descriptions

KyleKing:
I noticed a very minor nitpick in the documentation that I think is isolated to just these lines.

I have used `pyparsing` for a few years. Thanks for building and maintaining this library!


ptmcg:
Nice catch, thanks!

PR #454, opened 2022-12-02
Removed unused keyword var in lucene_grammar.py

lijenicol:
Found a variable in the `lucene_grammar.py` example which is not being used, I have removed it to keep things clean

### DIFF
--- a/examples/lucene_grammar.py
+++ b/examples/lucene_grammar.py
@@ -15,7 +15,6 @@ pp.ParserElement.enablePackrat()
 COLON, LBRACK, RBRACK, LBRACE, RBRACE, TILDE, CARAT = map(pp.Literal, ":[]{}~^")
 LPAR, RPAR = map(pp.Suppress, "()")
 and_, or_, not_, to_ = map(pp.CaselessKeyword, "AND OR NOT TO".split())
-keyword = and_ | or_ | not_ | to_
 
 expression = pp.Forward()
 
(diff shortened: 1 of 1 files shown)


ptmcg:
That's a nice catch, but it actually reflects a partial thought that was never completed.

In many parsers, I define a `keyword` expression so that the parser does not mistake a keyword for an identifier. This usually looks something like:

    valid_identifier = Combine(~keyword + Word(alphas, alphanums))

For the lucene parser, instead of guarding against mistaking a keyword for an identifier, we want to guard against mistaking a keyword for a term (so that we don't mistakenly think that "and" in the query means we want to search for the word "and").

The place for doing this would be in the assignment into the `term` expression. Since `term` is a Forward, we actually look for where the contents of `term` are defined using `<<=` operator.

```
term << (
        pp.Optional(field_name("field") + COLON)
        + (word_expr | string_expr | range_search | pp.Group(LPAR + expression + RPAR))
        + pp.Optional(boost)
    )
```

To follow this pattern, we would change it to:
```
term << (
        ~keyword
        + pp.Optional(field_name("field") + COLON)
        + (word_expr | string_expr | range_search | pp.Group(LPAR + expression + RPAR))
        + pp.Optional(boost)
    )
```

If you could update your PR to make this change instead of removing keyword, I'll be happy to merge it on in.

ptmcg:
Welcome to pyparsing! Please read some of the HOWTO information in the online docs, and in the Github project wiki.

PR #461, opened 2023-01-22
Update and rename invRegex.py to inv_regex.py

rjdbcm:
PEP8 update of invRegex example I converted for some experiments. Figured I'd make a PR for it.

-Ross


ptmcg:
Looks good, thanks!

PR #462, opened 2023-01-24
Simplify the `_trim_arity` function

heckad:
Use inspect module for getting arg count and then calculate the limit value instead of do brute force.


ptmcg:
Closing this PR - thanks again for submitting, but `trim_arity` is stable as-is.

PR #493, opened 2023-07-05
Limit error message overriding

ksunden:
This is a continuation from the conversation in matplotlib/matplotlib#26152.

I'm opening this here to further the conversation and so we can talk in slightly more concrete terms.

This is based on my last suggestion in that thread where I suggested Forward references without a name do not override the error message.

Either one of these factors or another error message override flag would be sufficient for matplotlib's purposes, so if you feel that this could be broader or would prefer to introduce a more explicit test, then feel free to say so and I will try to do so.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -4547,7 +4547,8 @@ class ParseElementEnhance(ParserElement):
             try:
                 return self.expr._parse(instring, loc, doActions, callPreParse=False)
             except ParseBaseException as pbe:
-                pbe.msg = self.errmsg
+                if not isinstance(self, Forward) or self.customName is not None:
+                    pbe.msg = self.errmsg
                 raise
         else:
             raise ParseException(instring, loc, "No expression defined", self)
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -9884,6 +9884,29 @@ class Test02_WithoutPackrat(ppt.TestParseResultsAsserts, TestCase):
                 msg=f"unexpected exception line ({exception_line!r})",
             )
 
+
+    def testForwardReferenceException(self):
+        token = pp.Forward()
+        num = pp.Word(pp.nums)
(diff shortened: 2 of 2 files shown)


ksunden:
Gentle ping @ptmcg, would like to move this conversation forward, happy to do more if that is wanted

ptmcg:
I haven't forgotten. On vacation this week, will try to sneak in a few minutes during quiet time, otherwise, next week.

ptmcg:
This change looks like you have narrowed the field down to being detected by a minimal condition, thanks! I'll look at it in more detail in my dev environment when I get back home in a few days.

Could you add some tests to test_unit.py for this please? Include some cases that are not strictly aligned with your usage (see how Forward is used elsewhere in the test cases or the examples). If you don't care to work with unittest (not everyone's favorite), just attach a small test file to this PR and I'll incorporate it into test_unit.py myself.

ksunden:
I've added a test which is a significantly pared down version of what we do in matplotlib (though does use the same mechanism for raising the more custom exception)

I do check as well, though, that if you set the name that the more default behavior returns.

Not fully sure what other test cases would be warranted, as this does trace both branches of the introduced conditional.

ksunden:
Gentle nudge on this one, we are preparing for a release and would like to fix up (or remove) pins on pyparsing/xfail tests based on the specific version of pyparsing (as everything _works_ other than having less helpful error messages, so may not want to wholly ban mpl from installing with pyparsing 3.1.0)

ptmcg:
Released today in pyparsing 3.1.1

ksunden:
Thanks, @ptmcg!

ptmcg:
I've added matplotlib-specific tests to the pyparsing test suite, cloned from your tests: https://github.com/pyparsing/pyparsing/blob/master/tests/test_matplotlib_cases.py

PR #510, opened 2023-09-01
infix_notation: typo checking type of lpar 

cpburnz:
It looks like you meant to check both *lpar* and *rpar* here to not be instances of *Suppress*.


ptmcg:
Thanks for the catch!

PR #518, opened 2023-11-07
ci: add Python 3.12 to test matrix

### LINKED ISSUE #517: Python 3.12 support
Is Python 3.12 supported or are there some known issues? I see the CI tests are not executed with Python 3.12, so it might be useful to add.

I'm asking because our code using on pyparsing is failing on Python 3.12 (see https://github.com/translate/translate/issues/5071). It might be as well bug on our side, but before digging into that, I'd like to know whether it is expected to work or not.

nijel:
Fixes #517


ptmcg:
Looks great, eager to try it out!

Issue #519, opened 2023-11-07
General code refactoring/denesting


InSyncWithFoo:
Recently I started using pyparsing for a project of mine. While doing debugging I noticed that the code is so deeply nested and thus extremely hard to read.

Please consider refactoring the code. I will be submitting some small PRs, but I'm afraid that I won't be able to modify those places with undocumented arguments and such.

ptmcg:
Released in 3.1.2 - thanks for the submission!

PR #520, opened 2023-11-07
Minor code refactoring/denesting for results.py

InSyncWithFoo:
This PR is a part of #519.

### DIFF
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -173,42 +173,52 @@ class ParseResults:
     ):
         self._tokdict: Dict[str, _ParseResultsWithOffset]
         self._modal = modal
-        if name is not None and name != "":
-            if isinstance(name, int):
-                name = str(name)
-            if not modal:
-                self._all_names = {name}
-            self._name = name
-            if toklist not in self._null_values:
-                if isinstance(toklist, (str_type, type)):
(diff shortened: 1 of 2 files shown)


ptmcg:
A number of unit tests fail with these changes. I'm grateful for your efforts, I'll see if I can help troubleshoot which specific changes are at issue.

ptmcg:
The changes to values() and items() cause the tests to fail. It is important that `__getitem__` is called to get the values and items for the keys, rather than just returning the values and items directly from tokdict. I can add some more explanatory comments on why they have to be the way they are, to make this clearer.

Also, please do not replace `not not expr` with `bool (expr)`. It is a micro-optimization, but anything I can do to shave cycles at parsing time is a plus, to help pyparsing's parse-time performance.

I like all the early exit code that gets rid of a fair bit of nesting, thanks!

jayaddison:
@InSyncWithFoo @ptmcg could you double-check the HTML file added in commit 865fccde7c956f95dc8544586a927cc8703de076?  It seems unrelated to the rest of the pull request, as far as I can tell.

ptmcg:
I'll revert it to its former state - thanks for the catch.

jayaddison:
Thanks both!

PR #522, opened 2023-11-08
Code refactoring for other files

InSyncWithFoo:
This PR is a part of #519.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -297,20 +297,21 @@ def _trim_arity(func, max_limit=3):
                 # re-raise TypeErrors if they did not come from our arity testing
                 if found_arity:
                     raise
+                else:
+                    tb = te.__traceback__
+                    frames = traceback.extract_tb(tb, limit=2)
+                    frame_summary = frames[-1]
+                    trim_arity_type_error = (
+                        [frame_summary[:2]][-1][:2] == pa_call_line_synth
+                    )
+                    del tb
 
-                tb = te.__traceback__
-                frames = traceback.extract_tb(tb, limit=2)
-                frame_summary = frames[-1]
-                trim_arity_type_error = (
-                    [frame_summary[:2]][-1][:2] == pa_call_line_synth
-                )
-                del tb
+                    if trim_arity_type_error:
+                        if limit < max_limit:
+                            limit += 1
+                            continue
 
-                if trim_arity_type_error and limit < max_limit:
-                    limit += 1
-                    continue
-
-                raise
+                    raise
     # fmt: on
 
     # copy func name to wrapper for sensible debug output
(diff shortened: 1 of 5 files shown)


pyparsing/unicode.py
```diff
@@ -102,14 +102,16 @@ def identbodychars(cls):
...
+        )
         return "".join(
             sorted(
                 set(
                     cls.identchars
                     + "0123456789·"
-                    + "".join(
-                        [c for c in cls._chars_for_ranges if ("_" + c).isidentifier()]
-                    )
+                    + "".join(identifier_chars)
```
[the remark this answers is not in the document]
ptmcg: You could just remove the []'s in line 111 and achieve the same result. IIRC, I intentionally made this a list comprehension instead of just a generator expression, since "".join can pre-allocate space when passed a list, but can't when passed a generator.

But since the whole reason we are building this internal string is to build a set, maybe the solution is to make this its own set using the generator you created or just a genexp, and '|' it with the smaller `set(cls.identchars + "0123456789·")`. Keep your change, but make `identifier_chars` a set.

pyparsing/core.py
```diff
@@ -297,21 +297,20 @@ def wrapper(*args):
                 # re-raise TypeErrors if they did not come from our arity testing
                 if found_arity:
                     raise
-                else:
```
ptmcg, line 300: Please remove all the changes to core.py from this PR. They are getting confused with the other.

pyparsing/helpers.py
```diff
@@ -95,15 +95,17 @@ def match_previous_literal(expr: ParserElement) -> ParserElement:
     rep = Forward()
 
     def copy_token_to_repeater(s, l, t):
-        if t:
```
ptmcg, line 98: Accepted.

pyparsing/helpers.py
```diff
@@ -787,7 +789,7 @@ def parseImpl(self, instring, loc, doActions=True):
     pa: typing.Optional[ParseAction]
     opExpr1: ParserElement
     opExpr2: ParserElement
-    for i, operDef in enumerate(op_list):
```
ptmcg, line 790: Accepted.

pyparsing/unicode.py
```diff
@@ -102,17 +102,11 @@ def identbodychars(cls):
         all characters in this range that are valid identifier body characters,
         plus the digits 0-9, and · (Unicode MIDDLE DOT)
         """
-        return "".join(
```
ptmcg, line 105: Accepted.

pyparsing/testing.py
```diff
@@ -180,49 +180,54 @@ def assertRunTestResults(
             """
             run_test_success, run_test_results = run_tests_report
 
-            if expected_parse_results is not None:
```
ptmcg, line 183: Accepted

PR #533, opened 2024-01-12
Use typing.Self where applicable

InSyncWithFoo:
`Self` should be preferred over `"ParserElement"` since the former is dedicated for exactly this purpose and may, in some cases, provides better insight than a non-discriminating `ParserElement`.

This change requires `typing_extensions` as a type-checking time dependency; it should not alter runtime behaviour, [unless someone](https://xkcd.com/1172/) inspects those type hints at runtime.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -17,6 +17,7 @@ from typing import (
     Tuple,
     Union,
     cast,
+    TYPE_CHECKING
 )
 from abc import ABC, abstractmethod
 from enum import Enum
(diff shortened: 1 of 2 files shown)


ptmcg:
At the moment, pyparsing has no required external run-time dependencies, unless you want to create railroad diagrams, which are typically only a dev-time requirement.

`Self` will be available from typing in 3.11 without any added dependencies. External dependencies that are version-sensitive are a major chore. I'm open to revisiting this if we get some compelling examples where `Self `will be significantly better.

PR #534, opened 2024-01-13
Avoid assigning an empty error message to delegated parse exception

### LINKED ISSUE #527: Custom error messages inside Group
Hello,

I have some custom error messages that used to work up to version 3.0.9, but on newer versions are no longer showing up.

Here is a minimal example that demonstrates the behaviour:

```python
import pyparsing as pp

def validate_function(original_string, loc, tokens):
    raise pp.ParseFatalException(original_string, loc, "This is custom fail message")

greeting_str = "hello world!"
greet = pp.Group(
    pp.Keyword("hello").set_parse_action(validate_function)
)

greet.parse_string(greeting_str)
```

On 3.0.9: 
```python
pyparsing.exceptions.ParseFatalException: This is custom fail message, found 'hello'  (at char 0), (line:1, col:1)
```

On version 3.1.1 I get:
```python
pyparsing.exceptions.ParseFatalException: , found 'hello'  (at char 0), (line:1, col:1)
```

This will only happen when ParseFatalException is raised for an expression inside a Group.

InSyncWithFoo:
Fixes #527.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -4510,7 +4510,8 @@ class ParseElementEnhance(ParserElement):
             return self.expr._parse(instring, loc, doActions, callPreParse=False)
         except ParseBaseException as pbe:
             if not isinstance(self, Forward) or self.customName is not None:
-                pbe.msg = self.errmsg
+                if self.errmsg:
+                    pbe.msg = self.errmsg
             raise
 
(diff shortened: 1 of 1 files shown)


lemenkov:
Fixes #527 for me. Also fixes eerimoq/asn1tools#167.
Hope it will hit upstream shortly.

PR #538, opened 2024-01-21
Accept floating-point NaN and Inf as fnumber

rwpeterson:
I was working on parsing numerical data and found that none of pyparsing's numerical expressions support the non-finite floating point values (NaN, +Inf, and -Inf).

Just based on the implementation, `ppc.fnumber` looks like a reasonable place to add support for this, since it's independent of the other numerical expressions and has the float conversion baked in. (Indeed, I initially thought it would accept these values due to the "float" part.) Essentially, this PR makes `fnumber` match anything that Python's `float` can itself parse. [cPython's parsing](https://github.com/python/cpython/blob/main/Python/pystrtod.c#L28) is fully case-insensitive and accepts (optionally signed) "NaN", "Inf", and "Infinity".

One note of caution would be if users somehow rely on `fnumber`'s failure to parse such constants as a way to deal with NaNs or infinities in their programs.

### DIFF
--- a/docs/HowToUsePyparsing.rst
+++ b/docs/HowToUsePyparsing.rst
@@ -1367,6 +1367,8 @@ Common string and token constants
 
 - ``common.fnumber`` - any numeric expression; parsed tokens are converted to float
 
+- ``common.ieee_float`` - any floating-point literal (int, real number, infinity, or NaN), returned as float
+
 - ``common.identifier`` - a programming identifier (follows Python's syntax convention of leading alpha or "_",
   followed by 0 or more alpha, num, or "_")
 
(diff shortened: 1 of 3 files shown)


ptmcg:
To avoid introducing a regression on `fnumber`, what do you think about adding separate expressions for nan's and inf's, so that apps that need them could use an expression like `numeric = fnumber | nan | inf`? Or create another version of `fnumber` that adds these - just need to come up with a decent name (one of the 2 hardest problems in CS).

rwpeterson:
I agree on keeping it separate from `fnumber`, and on the difficulty of a name :)

I factored it out into a separate `ieee_float`, since NaNs in particular are associated with the standard. Some other options for the name could be `float_literal`, `ieee754`, etc.

ptmcg:
`ieee_float` looks good to me, and I admit, I never would have come up with that. Thanks for submitting!

PR #540, opened 2024-02-14
Add codespell (config, workflow, pre-commit) to detect new typos, fix found typos

yarikoptic:
My humble contribution to make a world a better typo-free place

TODOs
- [ ] remove custom workflow if precommit finds my typo (in TMP)

there were prior commits committing typo fixes, even one using codespell to detect. Now it would be automated.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,23 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,10 @@
+name: Pre-commit
+on: [push, pull_request]
+jobs:
+  pre-commit:
+    name: Format
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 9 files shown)


ptmcg:
This looks interesting. Please don't check/modify pyparsing_archive.py though. The code that _should_ be checked is in the examples, tests, and pyparsing directories. I don't see where you define the .py files that should be checked.

yarikoptic:
By default it was by exclusion, not inclusion, i.e. it checks all text files (.py or not), and then we exclude some. It would be trivial to skip `pyparsing_archive.py` but are you sure should it be excluded?  

- prior commit to it 5353ccdd7026a7eeaa77029102f8c0043553ebd3 was exactly that -- fixing typos identified by codespell
- identified typos are legit

if it is not vendored code from somewhere else which might be subject to update and where typos should be fixed at the upstream project -- I do not see why it should not be fixed up.

Another interesting aspect is that I do not see you running precommit either through precommit "service" or github workflow since I did add pre-commit configuration, but it was only the codespell workflow I added which tripped on the injected typo.  Should I may be replace codespell workflow with some workflow which runs pre-commit?

ptmcg:
Never mind, I'm deleting the file. It was there as a security blanket for me while going through the breakup of the single-file release back during the pyparsing 2.x->3 transition, and I was still very unsure of myself that I might lose it out of git ignorance.

I can always get this from the 2.4.x branch (which I have locked in Github to prevent accidental loss or update to this code).

As for pre-commit, this is another case where I have been slow in adopting new tooling. It took me a while just to warm up to using black. Now I use pre-commit (with black, isort, flake8 and others) at work and rely on it heavily. If you want to open a PR that adds pre-commit as a Github workflow, that works great for me.

yarikoptic:
I have force pushed now with the pre-commit workflow -- please bless it to see if works ok

ptmcg:
Closing - will do better at spell-checking in future.

yarikoptic:
a worthy goal -- good luck!

PR #546, opened 2024-03-06
Refactor: remove one isinstance comparison from the runtime code path.

jayaddison:
This is a possible refactor resulting from [inspecting the differences between `v3.1.1` and `v3.1.2`](https://github.com/pyparsing/pyparsing/compare/3.1.1...3.1.2) - I believe that it should remove one conditional/comparison during at typical runtime evaluation without affecting the correctness of the code; and indenting of the source is not affected.

Arguably it does hide the validation-checks a little, by moving one of the `NotImplemented` return paths lower down the method body.

In particular, the `isinstance(other, tuple) and other[:1] == (Ellipsis,)` check and variable reassignments are, I believe, already handled by the code here:

https://github.com/pyparsing/pyparsing/blob/7d4bda2743ebc04f68d2594bc4fffc70cd65848f/pyparsing/core.py#L1518

(which should only be reachable [when `other` is indeed a `tuple`](https://github.com/pyparsing/pyparsing/blob/7d4bda2743ebc04f68d2594bc4fffc70cd65848f/pyparsing/core.py#L1512-L1517))

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1506,15 +1506,10 @@ class ParserElement(ABC):
         """
         if other is Ellipsis:
             other = (0, None)
-        elif isinstance(other, tuple) and other[:1] == (Ellipsis,):
-            other = ((0,) + other[1:] + (None,))[:2]
-
-        if not isinstance(other, (int, tuple)):
-            return NotImplemented
 
         if isinstance(other, int):
             minElements, optElements = other, 0
-        else:
+        elif isinstance(other, tuple):
             other = tuple(o if o is not Ellipsis else None for o in other)
             other = (other + (None, None))[:2]
             if other[0] is None:
@@ -1531,6 +1526,8 @@ class ParserElement(ABC):
                 optElements -= minElements
             else:
                 return NotImplemented
+        else:
+            return NotImplemented
 
         if minElements < 0:
             raise ValueError("cannot multiply ParserElement by negative value")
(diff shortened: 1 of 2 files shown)


ptmcg:
The updated test code is great.

As for the other code changes, I need to give them some thought. This code is already fairly obscure. The first part of normalizing `...` to the `(0, None)` tuple should then lead directly into the remaining code. If anything, I would refactor to make the code paths and decision logic more clear.

If you are considering performance in this change, that is a good thought, but here is my take on performance tuning in pyparsing: For performance issues, I mostly concern myself with performance that may occur in any of the `parseImpl `methods, or any supporting methods (like any code involved in creating `ParseExceptions` or `ParseResults`, which are natural products of the `parseImpl`s). These are the methods that get run when actually parsing. Any given line in `parseImpl `may get called many hundreds or thousands of times, so performance speedups there can accumulate into a measurable savings. Code involved in _creating_ the parse expressions is less critical to me (unless egregiously slow), as that is usually a one-time upfront task - since it is generally run only once or a few times (i.e., far fewer than the code in `parseImpl`), you really have to save a lot to make it worthwhile. For the code in this PR in particular, which is part of parser construction and not actual parsing, I would focus on readability and maintainability vs trying to eke out a few microseconds. (Performance might not even been part of your motivation here, but just wanted to share these thoughts.)

ptmcg:
Let's do this - resubmit your PR with those changes in mind.  Keep the new tests.  For the other code, refactor by:
- normalize any Ellipsis values (either standalone or as part of a tuple) to correct 0 or None values up front
- remove redundant Ellipsis-handling code after that
- rework obscure expressions like `other = (other + (None, None))[:2]`

jayaddison:
Will do - thanks @ptmcg!

PR #547, opened 2024-03-07
Refactor: attempt to simplify the multiply operator.

jayaddison:
* Reduce the amount of special-case logic required to handle the `Ellipsis` operator (somewhat equivalent/similar to `None` values -- translating to: unbound).
* Extract some parallels between the handling of "zero, one, or more" in the cases of unbounded-maximum or singleton repetitions.

cc @ptmcg

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1510,6 +1510,8 @@ class ParserElement(ABC):
             minElements, maxElements = (other + (None, None))[:2]
         elif other is Ellipsis:
             minElements, maxElements = other, None
+        else:
+            return NotImplemented
 
         if minElements in (Ellipsis, None):
             minElements = 0
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -1508,28 +1508,25 @@ class ParserElement(ABC):
             minElements, maxElements = other, other
         elif isinstance(other, tuple):
             minElements, maxElements = (other + (None, None))[:2]
+            minElements = 0 if minElements in (None, Ellipsis) else minElements
+            maxElements = None if maxElements in (None, Ellipsis) else maxElements
         elif other is Ellipsis:
-            minElements, maxElements = other, None
+            minElements, maxElements = 0, None
         else:
             return NotImplemented
 
-        if minElements in (Ellipsis, None):
-            minElements = 0
-        elif type(minElements) is not int:
-            return NotImplemented
-        elif minElements < 0:
+        if minElements < 0:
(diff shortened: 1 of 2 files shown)


pyparsing/core.py
```diff
@@ -1504,64 +1504,43 @@ def __mul__(self, other) -> "ParserElement":
...
-            other = ((0,) + other[1:] + (None,))[:2]
-
-        if not isinstance(other, (int, tuple)):
-            return NotImplemented
-
         if isinstance(other, int):
-            minElements, optElements = other, 0
+            minElements, maxElements = other, other
+        elif isinstance(other, tuple):
+            minElements, maxElements = (other + (None, None))[:2]
```
jayaddison, line 1510: This obscure-ish expression remains; I haven't yet been able to figure out a way to rewrite this that is more readable and type-safe.

jayaddison: It'd be nice to consider the `other` argument as if it were a token stream, and to handle `Ellipsis` as a special case there, consuming exactly two tokens, and with `None` emitted as a fallback if the stream is empty when reading the next token.

jayaddison: No, I can't think of a neater pattern for this unpacking assignment in the code at the moment, and won't make further changes to it.

ptmcg:
Not ignoring you, I have some work to catch up on tomorrow and over the weekend.

pyparsing/core.py
```diff
@@ -1504,64 +1504,43 @@ def __mul__(self, other) -> "ParserElement":
...
+            if minElements == 0:
+                return And([])
+            elif minElements == 1:
+                return self
             else:
-                ret = And([self] * minElements)
-        return ret
+                return And([self] * minElements)
+
+        return And([self] * minElements) + (Opt(self) * (maxElements - minElements))
```
jayaddison, line 1543: No hurry on code review; but from re-inspecting this myself, I'm beginning to wonder about this `Opt` construction and the effect this could have on the parse output.

In particular: perhaps it's better to have a chain of nested `Opt` nodes (recursive, instead of iterative as in this PR), because then matching an input against the parse tree can exit as soon as a single token _doesn't_ match -- whereas the `And[Opt(...), Opt(...), ...]` formulation here would be evaluated in full (I think?).

jayaddison:
@ptmcg I think I'll close this branch in the next few days, unless there's anything of use here (the test coverage perhaps?) - I don't think it makes any significant improvements, and that'd be one less PR to keep track of.

PR #549, opened 2024-03-08
Added a new example showcasing a recursive parser for parsing advance…

gialmisi:
…d mathematical expressions in infix notation.

The parser supports binary, unary, and variadic operators that can be customized by modifying the class variables. It therefore goes way beyond any existing examples for artithmetic expression parsers. It also showcases how regex can be utilized to improve the performance of a parser written utilzing pyparsing.

I wrote this parser for another project, where it was needed. I could not find a ready-made parser for my purposes in the examples, or the web in general. Therefore, I am making this pull request in hopes it gets accepted and hopefully lessens the burden of some other future user, who might have a similar need for a parser as I did. I left comments in the code that I made for myself when learning the library. I believe they add to the example, and therefore left them there.

### DIFF
--- a/examples/infix_math_parser.py
+++ b/examples/infix_math_parser.py
@@ -13,20 +13,18 @@ is parsed into
 [['Max', [[['Ln', ['x']], '+', ['Lb', [['Abs', ['y']]]]], ['Ceil', [[['Sqrt', ['garlic']], '*', 3]]], [['potato', '**', 2], '/', 4], [['Abs', ['cosmic']], '+', 10]]]]
 """
 
-from operator import or_
-from typing import ClassVar
-
 from pyparsing import (
     Forward,
     Group,
     Literal,
     ParserElement,
     Suppress,
-    delimitedList,
-    infixNotation,
+    DelimitedList,
+    infix_notation,
     one_of,
-    opAssoc,
+    OpAssoc,
     pyparsing_common,
+    ParseResults,
     Regex,
 )
 
@@ -39,7 +37,7 @@ class InfixExpressionParser:
 
     # Supported infix binary operators, i.e., '1+1'. The key is the notation of the operator in infix format,
     # and the value the notation in parsed format.
-    BINARY_OPERATORS: ClassVar[dict] = {
+    BINARY_OPERATORS: dict[str, str] = {
         "+": "Add",
         "-": "Subtract",
         "*": "Multiply",
@@ -49,7 +47,7 @@ class InfixExpressionParser:
 
     # Supported infix unary operators, i.e., 'Cos(90)'. The key is the notation of the operator in infix format,
     # and the value the notation in parsed format.
-    UNARY_OPERATORS: ClassVar[dict] = {
+    UNARY_OPERATORS: dict[str, str] = {
         "Cos": "Cos",
         "Sin": "Sin",
         "Tan": "Tan",
@@ -73,13 +71,12 @@ class InfixExpressionParser:
         "Sinh": "Sinh",
         "Tanh": "Tanh",
         "Rational": "Rational",
-        "-": "Negate",
     }
 
     # Supported infix variadic operators (operators that take one or more comma separated arguments),
     # i.e., 'Max(1,2, Cos(3)). The key is the notation of the operator in infix format,
     # and the value the notation in parsed format.
-    VARIADIC_OPERATORS: ClassVar[dict] = {"Max": "Max"}
+    VARIADIC_OPERATORS: dict[str, str] = {"Max": "Max"}
 
     def __init__(self):
         """A parser for infix notation, e.g., the human readable way of notating mathematical expressions.
@@ -94,8 +91,8 @@ class InfixExpressionParser:
         rparen = Suppress(")")
 
         # Define keywords (Note that binary operators must be defined manually)
-        symbols_variadic = set(InfixExpressionParser.VARIADIC_OPERATORS.keys())
-        symbols_unary = set(InfixExpressionParser.UNARY_OPERATORS.keys())
+        symbols_variadic = set(InfixExpressionParser.VARIADIC_OPERATORS)
+        symbols_unary = set(InfixExpressionParser.UNARY_OPERATORS)
 
         # Define binary operation symbols (this is the manual part)
         # If new binary operators are to be added, they must be defined here.
@@ -104,32 +101,29 @@ class InfixExpressionParser:
         plusop = one_of("+ -")
         expop = Literal("**")
 
-        # Dynamically create Keyword objects for variadric functions
+        # Dynamically create Keyword objects for variadic functions
         variadic_pattern = r"\b(" + f"{'|'.join([*symbols_variadic])}" + r")\b"
-        variadic_func_names = Regex(variadic_pattern)
+        variadic_func_names = Regex(variadic_pattern).set_name("variadic function")
 
         # Dynamically create Keyword objects for unary functions
         unary_pattern = r"\b(" + f"{'|'.join([*symbols_unary])}" + r")\b"
-        unary_func_names = Regex(unary_pattern)
+        unary_func_names = Regex(unary_pattern).set_name("unary function")
 
         # Define operands
-        integer = pyparsing_common.integer.set_parse_action(
-            pyparsing_common.convertToInteger
-        )
+        # Integers
+        integer = pyparsing_common.integer.set_name("integer")
 
         # Scientific notation
-        scientific = pyparsing_common.sci_real.set_parse_action(
-            pyparsing_common.convert_to_float
-        )
+        scientific = pyparsing_common.sci_real.set_name("float")
 
         # Complete regex pattern with exclusions and identifier pattern
         exclude = f"{'|'.join([*symbols_variadic, *symbols_unary])}"
         pattern = r"(?!\b(" + exclude + r")\b)(\b[a-zA-Z_][a-zA-Z0-9_]*\b)"
-        variable = Regex(pattern)
+        variable = Regex(pattern).set_name("variable")
 
         operands = variable | scientific | integer
 
-        # Forward declarations of variadric and unary function calls
+        # Forward declarations of variadic and unary function calls
         variadic_call = Forward()
         unary_call = Forward()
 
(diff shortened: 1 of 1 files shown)


ptmcg:
Looking over your parser this weekend, thanks for submitting it. In the meantime, please look at the plusminus package (https://github.com/pyparsing/plusminus). This package might do much of what you want, plus add evaluation as well as parsing.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+    expressions = [
+        "f_1 + f_2 - 1e-3",
+        "(x_1 + (x_2 * (c_1 + 3.3) / (x_3 - 2))) * 1.5",
+        "Max(Ln(x) + Lb(Abs(y)), Ceil(Sqrt(garlic) * 3), (potato ** 2) / 4, Abs(cosmic) + 10)",
+        "Max(Sqrt(Abs(x) + y ** 2), Lg(Max(cosmic, potato)), Ceil(Tanh(x) + Arctan(garlic)))",
+        "((garlic**3 - 2**Lb(cosmic)) + Ln(x**2 + 1)) / (Sqrt(Square(y) + LogOnePlus(potato + 3.1)))",
+    ]
+
+    for expr in expressions:
```
ptmcg, line 189: Instead of doing your own loop (and handling exceptions), try `infix_parser.expn.run_tests(expressions)`

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+from typing import ClassVar
+
+from pyparsing import (
+    Forward,
+    Group,
+    Literal,
+    ParserElement,
+    Suppress,
+    delimitedList,
+    infixNotation,
```
ptmcg, line 26: Thanks for using the new PEP8 names where possible, you missed a couple (`infix_notation`, `DelimitedList`).

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+    one_of,
+    opAssoc,
```
ptmcg, line 28: `OpAssoc` is preferred over `opAssoc`

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        "Arccosh": "Arccosh",
+        "Arcsin": "Arcsin",
+        "Arcsinh": "Arcsinh",
+        "Arctan": "Arctan",
+        "Arctanh": "Arctanh",
+        "Cosh": "Cosh",
+        "Sinh": "Sinh",
+        "Tanh": "Tanh",
+        "Rational": "Rational",
+        "-": "Negate",
```
ptmcg, line 76: I don't think this will ever evaluate as "Negate", since "-" is a signop.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        # Dynamically create Keyword objects for variadric functions
+        variadic_pattern = r"\b(" + f"{'|'.join([*symbols_variadic])}" + r")\b"
+        variadic_func_names = Regex(variadic_pattern)
+
+        # Dynamically create Keyword objects for unary functions
+        unary_pattern = r"\b(" + f"{'|'.join([*symbols_unary])}" + r")\b"
+        unary_func_names = Regex(unary_pattern)
+
+        # Define operands
+        integer = pyparsing_common.integer.set_parse_action(
```
ptmcg, line 116: Adding parse actions to integer and scientific is not necessary, since the expressions in pyparsing.common already have these same parse actions.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+Examples of parsing:
+
+The expression "f_1 + f_2 - 1e-3" is parsed into [['f_1', '+', 'f_2', '-', 0.001]]
+
+The expression "Max(Ln(x) + Lb(Abs(y)), Ceil(Sqrt(garlic) * 3), (potato ** 2) / 4, Abs(cosmic) + 10)"
+is parsed into
+[['Max', [[['Ln', ['x']], '+', ['Lb', [['Abs', ['y']]]]], ['Ceil', [[['Sqrt', ['garlic']], '*', 3]]], [['potato', '**', 2], '/', 4], [['Abs', ['cosmic']], '+', 10]]]]
+"""
+
+from operator import or_
```
ptmcg, line 16: `or_` is not used anywhere

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        "+": "Add",
+        "-": "Subtract",
+        "*": "Multiply",
+        "/": "Divide",
+        "**": "Power",
+    }
+
+    # Supported infix unary operators, i.e., 'Cos(90)'. The key is the notation of the operator in infix format,
+    # and the value the notation in parsed format.
+    UNARY_OPERATORS: ClassVar[dict] = {
```
ptmcg, line 52: Feels a little misleading to refer to UNARY and VARIADIC as "operators". I see these more as "function names", which you later use to compose `unary_call` and `variadic_call`.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        variadic_call <<= Group(
+            variadic_func_names + lparen + Group(delimitedList(infix_expn)) + rparen
+        )
+        unary_call <<= Group(unary_func_names + lparen + Group(infix_expn) + rparen)
+
+        self.expn = infix_expn
+
+        # The infix operations do not need to be in this list because they are handled by infixNotation() above.
+        # If new binary operations are to be added, they must be updated in the infixNotation() call (the list).
+        self.reserved_symbols: set[str] = symbols_unary | symbols_variadic
```
ptmcg, line 164: Does not seem to be used anywhere.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # It is assumed that the dicts in the three class variables have unique keys.
+        self.operator_mapping = {
```
ptmcg, line 167: Does not seem to be used anywhere.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+            **InfixExpressionParser.BINARY_OPERATORS,
+            **InfixExpressionParser.UNARY_OPERATORS,
+            **InfixExpressionParser.VARIADIC_OPERATORS,
+        }
+
+    def parse(self, str_expr: str) -> list:
+        """Parse a string expression into a list."""
+        return self.expn.parse_string(str_expr, parse_all=True)
```
ptmcg, line 175: This does not return a list, but a `pyparsing.ParseResults `object. You can convert it to a list using `.as_list()`.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Scientific notation
+        scientific = pyparsing_common.sci_real.set_parse_action(
+            pyparsing_common.convert_to_float
+        )
+
+        # Complete regex pattern with exclusions and identifier pattern
+        exclude = f"{'|'.join([*symbols_variadic, *symbols_unary])}"
+        pattern = r"(?!\b(" + exclude + r")\b)(\b[a-zA-Z_][a-zA-Z0-9_]*\b)"
+        variable = Regex(pattern)
```
ptmcg, line 128: Adding `.set_name("variable")` will make your error messages nicer looking. Likewise for `integer`, `scientific`, and others.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+# Enable Packrat for better performance in recursive parsing
+ParserElement.enablePackrat(None)
+
+
+class InfixExpressionParser:
+    """A class for defining an infix notation parsers."""
+
+    # Supported infix binary operators, i.e., '1+1'. The key is the notation of the operator in infix format,
+    # and the value the notation in parsed format.
+    BINARY_OPERATORS: ClassVar[dict] = {
```
ptmcg, line 42: I like the type annotations, but I don't think the `ClassVar` is that useful here. The primary purpose behind `ClassVar` is to signal type-checkers to flag cases where an instance defines its own instance variable by the same name (see https://docs.python.org/3/library/typing.html#typing.ClassVar).  `dict[str, str]` would probably be more appropriate, or even define your own TypeAlias for `dict[str, str]`, and use it for all these mapping vars.

examples/infix_math_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        "Cos(2 + f_1) - 7.2 + Max(f_2, -f_3)" is parsed to the list:
+        ['Cos', [[2, '+', 'f_1']]], '-', 7.2, '+', ['Max', ['f_2', ['-', 'f_3']].
+
+        """
+        # Scope limiters
+        lparen = Suppress("(")
+        rparen = Suppress(")")
+
+        # Define keywords (Note that binary operators must be defined manually)
+        symbols_variadic = set(InfixExpressionParser.VARIADIC_OPERATORS.keys())
```
ptmcg, line 97: Calling keys() is unnecessary here, since iterating over a dict gives an iterator over its keys.

ptmcg (commented):
Nice example, glad to see others making use of `infix_notation` like this.

examples/infix_math_parser.py
gialmisi: I was not aware of this. Thanks!

examples/infix_math_parser.py
gialmisi: This is true. This is an artifact from the application I have worked and I extracted this example from.

examples/infix_math_parser.py
gialmisi: This is nice to know, I guess this also improves performance a little bit to not add explicit parse actions? Anyhow, I removed them from the example.

examples/infix_math_parser.py
gialmisi: I believe I added them to all relevant patterns.

gialmisi:
> Looking over your parser this weekend, thanks for submitting it. In the meantime, please look at the plusminus package (https://github.com/pyparsing/plusminus). This package might do much of what you want, plus add evaluation as well as parsing.

Thanks, I wish I had found this before, might have been useful!

As for the other comments, I have addressed them and updated the PR. I always appreciate someone going over code I have written and giving feedback, especially with this level of care. Thanks!

The diffs are somehow messed up in the PR. I should check my IDE settings. Sorry for this. I would be happy if this example is included in pyparsing, but if its contribution to the framework is not enough to warrant a pull, that is fine as well.

Yedfree (approved):

ptmcg:
I was going to suggest that the unary_func wasn't necessary, since the variadic_func will match a single argument. But this does work to do verification of the number of arguments at parse time for specific functions. Thanks for being patient on this PR.

Issue #564, opened 2024-08-25
pyparsing 3.1.3 fails on Python 3.6


felixfontein:
The class `re.Pattern` only exists since Python 3.7, but is used in https://github.com/pyparsing/pyparsing/blob/master/pyparsing/core.py#L3060.

```
  Traceback (most recent call last):
    File "/home/runner/work/antsibull-docs-parser/antsibull-docs-parser/antsibull-docs-parser/.nox/test-3-6/lib/python3.6/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
      main()
    File "/home/runner/work/antsibull-docs-parser/antsibull-docs-parser/antsibull-docs-parser/.nox/test-3-6/lib/python3.6/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/runner/work/antsibull-docs-parser/antsibull-docs-parser/antsibull-docs-parser/.nox/test-3-6/lib/python3.6/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 160, in prepare_metadata_for_build_wheel
      whl_basename = backend.build_wheel(metadata_directory, config_settings)
    File "/tmp/pip-build-env-1qhbq5lc/overlay/lib/python3.6/site-packages/hatchling/build.py", line 41, in build_wheel
      return os.path.basename(next(builder.build(wheel_directory, ['standard'])))
    File "/tmp/pip-build-env-1qhbq5lc/overlay/lib/python3.6/site-packages/hatchling/builders/plugin/interface.py", line 76, in build
      self.metadata.core.validate_fields()
    File "/tmp/pip-build-env-1qhbq5lc/overlay/lib/python3.6/site-packages/hatchling/metadata/core.py", line 1115, in validate_fields
      getattr(self, attribute)
    File "/tmp/pip-build-env-1qhbq5lc/overlay/lib/python3.6/site-packages/hatchling/metadata/core.py", line 1014, in dependencies
      self._dependencies = sorted(self.dependencies_complex)
    File "/tmp/pip-build-env-1qhbq5lc/overlay/lib/python3.6/site-packages/hatchling/metadata/core.py", line 971, in dependencies_complex
      from packaging.requirements import InvalidRequirement, Requirement
[...]

PR #565, opened 2024-08-25
[3.1.3] Revert 3bb8076 for Python 3.6

hswong3i:
Revert "Some PEP8 naming and type annotations; added some test cases for string->regex in Word"

This reverts commit 3bb8076ffae804f398f14af0d887e131c9ffb312.

Fixes https://github.com/pyparsing/pyparsing/issues/564

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -470,7 +470,7 @@ class ParserElement(ABC):
         self.streamlined = False
         # optimize exception handling for subclasses that don't advance parse index
         self.mayIndexError = True
-        self.errmsg: Union[str, None] = ""
+        self.errmsg = ""
         # mark results names as modal (report only last) or cumulative (list all)
         self.modalResults = True
         # custom debug actions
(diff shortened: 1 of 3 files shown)


ptmcg:
There is no need to revert this entire commit. The only change that breaks Python 3.6 compatibility is the use of `re.Pattern`. I would like to keep some form of type annotation here, replacing the use of `re.Pattern` which was introduced in Python 3.7. If you can change this PR to just remove (or comment out) the reference to `re.Pattern` in core.py, I'll accept and merge your PR, and then follow up later with either a compatibility version, or just leave the commented-out code in place until I release pyparsing 3.2.0 (which will drop support for Python 3.6 and 3.7, which is kind of overdue).

Thanks for submitting!

ptmcg:
There is another PR submitted that fixes this issue, and also updates the CI file - being more complete, I'm going to use that one, and close this one.

Thanks for using pyparsing!

hswong3i:
@ptmcg thank you very much for 3.1.4 release; this PR is primary for my https://build.opensuse.org/package/show/home:alvistack/pyparsing-pyparsing-3.1.3 packaging, so only keep it working as simple as possible by revert commit ;-)

PR #571, opened 2024-09-22
Add found string as property

OdysseasKr:
A bit of a silly change, feel free to close if you don't accept these.

I have moved the logic of the "found" string out of the `__str__` magic method. This allows the user to generate prettier messages from the exception. I also think it's a bit cleaner to avoid parsing logic in `__str__`

## Example
From the following exception

```Python
In [2]: Keyword("KEY1").parse_string("KEY2", parseAll=True)
---------------------------------------------------------------------------
ParseException                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 Keyword("KEY1").parse_string("KEY2", parseAll=True)

File ~/Documents/pyparsing/pyparsing/core.py:1197, in ParserElement.parse_string(self, instring, parse_all, parseAll)
   1194         raise
   1195     else:
   1196         # catch and re-raise exception from here, clearing out pyparsing internal stack trace
-> 1197         raise exc.with_traceback(None)
   1198 else:
   1199     return tokens

ParseException: Expected Keyword 'KEY1', found 'KEY2'  (at char 0), (line:1, col:1)
```
I would like to generate a message like this

```
1:1 Expected Keyword 'KEY1', found 'KEY2' 
```

With the existing implementation the user would have to call  `__str__` and then remove the character, line and column information.

### DIFF
--- a/pyparsing/exceptions.py
+++ b/pyparsing/exceptions.py
@@ -5,15 +5,16 @@ import copy
 import re
 import sys
 import typing
+from functools import cached_property
 
+from .unicode import pyparsing_unicode as ppu
 from .util import (
+    _collapse_string_to_ranges,
     col,
     line,
     lineno,
-    _collapse_string_to_ranges,
     replaced_by_pep8,
 )
-from .unicode import pyparsing_unicode as ppu
 
 
 class _ExceptionWordUnicodeSet(
@@ -134,35 +135,35 @@ class ParseBaseException(Exception):
         """
         return cls(pe.pstr, pe.loc, pe.msg, pe.parser_element)
 
-    @property
+    @cached_property
     def line(self) -> str:
         """
         Return the line of text where the exception occurred.
         """
         return line(self.loc, self.pstr)
 
-    @property
+    @cached_property
     def lineno(self) -> int:
         """
         Return the 1-based line number of text where the exception occurred.
         """
         return lineno(self.loc, self.pstr)
 
-    @property
+    @cached_property
     def col(self) -> int:
         """
         Return the 1-based column on the line of text where the exception occurred.
         """
         return col(self.loc, self.pstr)
 
-    @property
+    @cached_property
     def column(self) -> int:
         """
         Return the 1-based column on the line of text where the exception occurred.
         """
         return col(self.loc, self.pstr)
 
-    @property
+    @cached_property
     def found(self) -> str:
         if not self.pstr:
             return ""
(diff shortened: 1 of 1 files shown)


ptmcg:
I'll give it a look. There are a few very popular projects that use pyparsing and rely on the specific format of exception messages, so I try to keep these messages from changing if I can help it.

Adding a new property should not be a problem, and with that you might be able to monkeypatch `ParseBaseException.__str__` to reformat it to your liking.

And those projects might also benefit from this new property, so there is that possibililty as well (though pyparsing elements of most projects are often in the category of "that one guy wrote this 10 years ago, so let's not ever change it").

pyparsing/exceptions.py
```diff
@@ -162,6 +162,23 @@ def column(self) -> int:
...
+
+        # pull out next word at error location
+        found_match = _exception_word_extractor.match(self.pstr, self.loc)
+        if found_match is not None:
+            found = found_match.group(0)
+        else:
+            found = self.pstr[self.loc : self.loc + 1]
+
+        return repr(found).replace(r"\\", "\\")
+
```
ptmcg, line 181: This looks good, but I think I want to make some minor refactoring changes:
- change this and all the other properties to `cached_property`, imported from functools
- move the formatting that is done in `__str__` to a new method, `formatted_message` (see below)
- change `__str__` to just return `self.formatted_message()`

```python
    def formatted_message(self):
        found_phrase = f", found {self.found}" if self.found else ""
        return f"{self.msg}{found_phrase}  (at char {self.loc}), (line:{self.lineno}, col:{self.column})"
```

(I liked `found_phrase` over `foundstr`, as I advance in my pyparsing support years, I'm trying to make variable names more readable and intuitive.)

This will allow you to format your own custom exception messages, like this (inserted at the beginning of your program, to monkeypatch the default `formatted_message` function).

```python
from pyparsing import Keyword, ParseBaseException

def custom_exception_message(exc):
    found_phrase = f", found {exc.found}" if exc.found else ""
    return f"{exc.lineno}:{exc.column} {exc.msg}{found_phrase}"

ParseBaseException.formatted_message = custom_exception_message

Keyword("KEY1").parse_string("KEY2", parseAll=True)
```

ptmcg:
Ok, I think this would make for a nice change, and there are likely some (slight) performance improvements we can do, too.  See my review comment.

ptmcg:
Or if you'd rather, I can accept your PR as-is, and then make these adjustments afterward (and also add tests and CHANGES note).

OdysseasKr:
Make sense, should have used cached_property for this. I will make the changes today!

PR #584, opened 2024-11-13
Improve performance of cpp_style_comment

gerlero:
Improve performance of the `cpp_style_comment` helper by rewriting it as a single `Regex`.

This change leads to very significant performance improvements when parsing grammars that use `.ignore(cpp_style_comment)`.


ptmcg:
This form dates back to 2016, when an issue with [regex catastrophic backtracking was reported](https://github.com/pyparsing/wikispaces_archive/blob/master/discussion/all_wiki_discussion_toc_2016.md#2016-04-28-085405---webpentest---pypasing-hangs-on-parsing-dblquotedstring) against the quoted string expressions. At that time, I assumed that this issue also applied to the C and C++ comment expressions as well, but after testing today, it does not seem to show the same bad behavior. Thanks for chasing this down, I'll merge this change and add the unit tests around the regex backtracking issue (there were already some, but I'm going to clean/tighten them up). And also a CHANGES note.

PR #586, opened 2024-11-23
Use non-capturing groups in common.fnumber and common.ieee_float

gerlero:
Non-capturing groups are faster to parse and IMO show the intent better. Also, the `(?i:...)` inline flag in this format does not need to be at the beginning of the regular expression, allowing for the pattern to be easily reused as part of other custom `Regex`es (e.g. by extracting it as `pyparsing.common.ieee_float.re.pattern`—not sure you want to encourage this, but TBF I'm doing something like this in my own project that uses this library).


ptmcg:
Looks good, thanks!

PR #589, opened 2024-12-14
Fix a typo and two unused names

dalps:
I think it's appropriate to use the names you've defined for the operators "!" nad "^" in place of their literals.


ptmcg:
Thanks for these updates, merging.

PR #595, opened 2025-01-08
Docs: Update ParseException's docstring example

ferdnyc:
This is incredibly minor, but... the example output in `ParseException`'s docstring had gotten out-of-date / at odds with reality. Fixed.


ptmcg:
LGTM - if you are interested there are a bunch more of these over in core.py

PR #606, opened 2025-06-05
QuotedString: Add caution about param interactions

ferdnyc:
Looking at adjusting the pydot GraphViz parser's handling of quoted strings, I was rather surprised to discover an unexpected interaction in the `QuotedString` parameters:

As implemented in `QuotedString`, `convert_whitespace_escapes` **has no effect** unless `unquote_results` is set to `True`. 

This PR adds a Caution admonition to the API documentation, advising users of this restriction.


ptmcg:
Thanks for digging in to this. Will push out in a point release this month.

PR #607, opened 2025-06-25
Docs: A raft of cleanup & style changes

ferdnyc:
This PR represents a set of tweaks to the project documentation, mostly focused around usability and presentation of information. A lot of it is obviously subjective, and while I personally thing each change is an improvement, I'm not married to any of it and am more than willing to take feedback / remove anything that's not wanted. It's all in very granular commits specifically so that it can easily be adjusted.

### What's here
(Formatted as a checklist, so we can mark off anything to be removed/changed)

- [ ] Turned all of the 3.0.0 "Deprecated" messages into real `.. deprecated::` markup
- [ ] Added `.. versionchanged::` and `.. versionadded::` markup for most of the items in the "what's new in 3.1.0 / 3.2.0" documents (some items don't easily translate into API documentation)
- [ ] Added custom CSS to style the change messages for visibility. It makes use of the relatively-new CSS `:has()` selector to push the `.. deprecated::` styling up to the containing class/function/method/etc. level, so that anything marked deprecated has a red border line running all the way down its left side for maximum visibility. You'll see in screenshots.
   (After some deliberation, I decided not to do the same with added/changed notes, because those won't be "recent" forever and as they age they deserve less prominence. Deprecated features warrant **greater** prominence, if anything, as time passes until they're eventually removed.)
- [ ] Eliminated the unnecessary `docs/modules.rst` document, and made `docs/pyparsing.rst` the top-level index for its own documentation (by putting a hidden `.. toctree::` at the bottom containing only `self`).
- [ ] Moved the documentation from a couple of classes' `__init__`s to the class itself, and removed the `__init__` signatures from the docs entirely, since the construction signatures are shown on the class itself and the `__init__`s become totally redundant, with the documentation moved.
- [ ] Removed the `.. sectnum::`s from the documentation pages, because Sphinx says not to use them, and with good reason -- they're totally broken. Every document gets numbered "1" in the TOC. There are apparently ways to make the sections properly numbered, but `sectnum` isn't it. (I personally don't think it's necessary.)

#### Sphinx configuration changes
- [ ] Leave class/module parents out of the TOC, for readability and to keep them from overflowing into the body text
- [ ] Break longer call signatures (over 100 characters, as currently configured) at each argument for "better" readability; it's still not great but better than one long, sometimes-four-or-five-wrapped-lines continuous signature string.
- [ ] Leave resolvable type names unexpanded in the signatures, which affects some signatures in small ways. (For example, the default `max_matches` argument for `ParserElement.scan_string()` is now listed as `_MAX_INT` instead of `9223372036854775807`.)

#### autodoc configuration changes

- [ ] I changed the `.. automodule::` in `pyparsing.rst` like so:

##### Before
```rst
.. automodule:: pyparsing
    :members:
    :special-members:
    :show-inheritance:
```

##### After
```rst
.. automodule:: pyparsing
    :members:
    :undoc-members:
    :special-members: __add__,__sub__,__div__,__mul__,__and__,__or__,__xor__,__call__,__weakref__,__str__
    :exclude-members: __init__,__repl__,parseImpl,parseImpl_regex,parseImplAsGroupList,parseImplAsMatch,postParse,preParse
    :show-inheritance:
```

Which, obviously, greatly affects the contents of the documentation. But I felt there were a number of undocumented members that were worth showing signatures for, and many special members that _weren't_, so explictly selecting the ones we actually want seemed more productive.

It is a lot of configuration, though, and isn't maintenance-free. (For example, after I added a `.. versionchanged::` to the `__str__` method for `ParseBaseException`, I had to go back and add `__str__` to the `:special-members:` list to get it to show up in the docs.)

So, even more than everything else, explicitly up for debate / open to revision.

I'll follow up with comment(s) showing some side-by-side screenshots, this opening message has gotten  long enough.

### DIFF
--- a/docs/HowToUsePyparsing.rst
+++ b/docs/HowToUsePyparsing.rst
@@ -19,8 +19,6 @@ Using the pyparsing module
     expressions, processing custom application language commands, or
     extracting data from formatted reports.
 
-.. sectnum::    :depth: 4
-
 .. contents::   :depth: 4
 
 Note: While this content is still valid, there are more detailed
--- a/docs/_static/pyparsing.css
+++ b/docs/_static/pyparsing.css
@@ -1,0 +1,44 @@
+/* Deprecated get a red border spanning the entire length of the doc
+ * (class, function, etc.), and the message itself has a faint red
+ * background shading, but no border. */
+dl.py:where(.class,.exception,.method,.function):has(> dd > div.deprecated)
+{
+    margin-inline-start: -0.6rem;
+    border-inline-start: 0.4rem solid #f00;
+    padding: 0 0.6rem 0 0.2rem;
+}
+
+span.deprecated {
+  background-color: #fee;
+  font-weight: bold;
+  text-decoration: #f00 underline;
+  padding: 0 0.5rem;
+}
+
+/* Added and changed get a blue or orange (respectively) border next to
+ * the message only, plus a light background shade of the same color
+ * (again, only on the message, not the rest of the doc). */
+div.versionadded, div.versionchanged
+{
+    border-inline-start: 0.4rem solid transparent;
+}
+
+div.versionchanged p,
+div.versionadded p,
+span.versionmodified {
+  line-height: initial;
+  padding-bottom: 0.2rem;
+  padding-top: 0.2rem;
+}
+span.versionmodified {
+  padding-inline-start: 0.5rem;
+  padding-inline-end: 0.2rem;
+  margin-inline-end: 0.5rem;
+  line-height: 130%;
+}
+
+div.versionadded { border-color: #2d67f3; }
+div.versionadded span.added { background-color: #d1e5ff; }
+
+div.versionchanged { border-color: #ff9800; }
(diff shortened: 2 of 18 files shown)


ferdnyc:
@ptmcg 

Wow, thanks very much!

We probably _should_ iterate a bit on my selections of `:special-members:` and `:excluded-members:`  for the autodocs, though.

Like, for example, I included `__weakrefs__` in the visible `:special-members:` purely because it was one of the symbols that the existing documentation showed, and I was making that decision at a point where I'd already changed so much, that I felt the need to be really conservative and minimize the changes that resulted from **those** decisions.

But I really don't think there's any value in `__weakrefs__` (which always has the default, Python-standard-library docstring, so nothing of value) showing up in the  docs. I'd always sort of planned to fold like a house of cards the moment someone questioned its inclusion. Except that never happened! :laughing: 

But I'm planning to open a follow-up issue about documentation anyway (in fact, I came here to do that right now), so that conversation can happen over there.

PR #609, opened 2025-06-30
ReadTheDocs: Pin sphinx to versions < 8.2

ferdnyc:
Also, explicitly install pyparsing with the `diagrams` extra, to ensure that railroad and jinja are available during docs build. (May be needed for doctests, API references.)


ptmcg:
Will give it a shot, thanks for chasing down the details.

ferdnyc:
🫡

I really need to make the time to either find or create a Sphinx bug report about this change, discover whether it was intentional or not, and see if there's some way to restore the old behavior via some configuration or syntax.

It'd be nice to _be able_ to eventually upgrade, staying pinned to Sphinx 8.1.x _forever_ doesn't seem like a workable long-term strategy.

PR #611, opened 2025-07-06
Add CONTRIBUTING doc to documentation

ferdnyc:
I'm planning to add a section to the `CONTRIBUTING.md` file about writing doctests for Pyparsing, and as I was doing that thought that it would make sense to make it part of the official documentation as well. So, this PR does that, which involves:

- Enable the `myst_parser` Python package as a Sphinx extension, which enables it to include MarkDown files in the documentation.
- Add `myst_parser` to `docs/requirements.txt` so that it's installed by ReadTheDocs and anyone else looking to build documentation.
- Symlink `CONTRIBUTING.md` into the `docs/` dir (this needs to be tested against ReadTheDocs; it may or may not work)
- Add `CONTRIBUTING` to the `index.rst` toctree.
- Update the title from "CONTRIBUTING" to "Contributing to Pyparsing".

That way, the content is included in the project documentation:

image

(Assuming ReadTheDocs doesn't balk at the symlink to the parent directory. It'll _proooobably_ be fine.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# CONTRIBUTING
+# Contributing to Pyparsing
 
 Thank you for your interest in working on pyparsing! Pyparsing has become a popular module for creating simple
 text parsing and data scraping applications. It has been incorporated in several widely-used packages, and is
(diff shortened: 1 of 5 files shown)


ferdnyc:
(Side note, that doc could use some updating. For example, it currently claims that Pyparsing supports Python 3.6.8+.)

PR #635, opened 2026-04-02
Fix as_dict() returning [] instead of {} for empty Dict results

nuglifeleoji:
## Problem (fixes #632)

When a named result wraps an empty `Dict(ZeroOrMore(...))` expression, `as_dict()` serializes the inner empty `ParseResults` as `[]` instead of `{}`.

```python
import pyparsing as pp

parser1 = pp.Dict(pp.ZeroOrMore(pp.Group(pp.Word(pp.alphas) + pp.Word(pp.alphas))))
parser2 = parser1.copy()('fubar')

parser1.parse_string('foo bar').as_dict()  # {'foo': 'bar'}  ✓
parser2.parse_string('foo bar').as_dict()  # {'fubar': {'foo': 'bar'}}  ✓

parser1.parse_string('').as_dict()  # {}  ✓
parser2.parse_string('').as_dict()  # {'fubar': []}  ✗  should be {'fubar': {}}
```

## Root Cause

The `to_item()` helper inside `as_dict()` uses `obj.haskeys()` to decide whether to serialize a nested `ParseResults` as a dict or a list. For an **empty** `Dict` result, `_tokdict` is empty (no entries were matched), so `haskeys()` returns `False`, causing the list path to be taken and returning `[]`.

## Fix

Add a lightweight `_is_dict_context: bool` flag to `ParseResults` (default `False`). `Dict.postParse` sets this flag to `True` on every tokenlist it processes — including empty ones. `to_item()` then checks `obj.haskeys() or obj._is_dict_context` so that empty dict contexts serialize as `{}`.

This preserves the existing behaviour for all other empty results (e.g. `ZeroOrMore(val)("values")` returning `[]` when empty), since those do not go through `Dict.postParse`.

## Tests

All 1977 existing tests pass. Verified manual reproduction from the issue is fixed.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -6627,6 +6627,11 @@ class Dict(TokenConverter):
                 else:
                     tokenlist[ikey] = _ParseResultsWithOffset(dictvalue[0], i)
 
+        # Mark the tokenlist so that as_dict() knows to serialize an empty
+        # result as {} rather than [], preserving dict semantics even when
+        # there are no matched entries.
+        tokenlist._is_dict_context = True
+
         if self._asPythonDict:
             return [tokenlist.as_dict()] if self.resultsName else tokenlist.as_dict()
 
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -105,6 +105,7 @@ class ParseResults:
         "_modal",
         "_toklist",
         "_tokdict",
+        "_is_dict_context",
     )
 
     class List(list):
@@ -185,6 +186,7 @@ class ParseResults:
         self._name = None
         self._parent = None
         self._all_names = set()
+        self._is_dict_context = False
 
         if toklist is None:
             self._toklist = []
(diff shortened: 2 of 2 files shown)


ptmcg:
Passing existing unit tests shows that you haven't broken anything, but doesn't prove that you've fixed anything either.

There is a bug in `__setstate__()`, fails to preserve value of `_is_dict_context`.

Add unit tests to verify fixed behavior, and persistence across pickle/unpickle.

Note to @nuglifeleoji : massive submissions of PRs to a multitude of repos marks you as a suspicious user. This is the kind of activity that a bad actor would employ to lay the groundwork for supply-chain attacks. Don't submit any more PRs to this project.

ptmcg:
If there is no further work on this PR in the next week, I'll close it and move forward with a more complete implementation.

ptmcg:
Adopted submitted changes as partial solution, added remaining pickle support and unit tests

PR #636, opened 2026-04-08
Fix match_previous_expr with leave_whitespace(recursive=True) (#633)

fcardosojr18:
### Problem
`match_previous_expr` fails when `leave_whitespace(recursive=True)` is used.

Example:
```python
first = pp.Char(pp.nums)
second = pp.match_previous_expr(first)
expr = first + ":" + second
expr.leave_whitespace(recursive=True)

list(expr.scan_string("1:2"))  # incorrectly matches

This should not match, the first and second values are different

**Problem**
leave_whitespace(recursive=True) replaces the child expressions with copies
self.exprs = [e.copy() for e in self.exprs]

match_previous_expr attatches the validation parse action to the original red(forward) object
rep.set_parse_action()

**FIX**
Attatch the validation parse action to e2 instead of rep so that the behavior is preserved even if the parser tree is copied

Tests
Added regressions tests in tests/test_issue_633.py

- an initially failing test case with recursive=True
- A control test with no whitespace modification
- additional test for when recursive=False
All tests including the original full test suite Pass

###Note
This is my first open source contribution as a student. I'd really appreciate any feedback, Thanks!

### CODE REVIEW

File: pyparsing/helpers.py [hunk 1]
```diff
@@ -165,8 +165,7 @@ def must_match_these_tokens(s, l, t):
                 raise ParseException(
                     s, l, f"Expected {matchTokens}, found{theseTokens}"
                 )
-
-        rep.set_parse_action(must_match_these_tokens, call_during_try=True)
+        e2.set_parse_action(must_match_these_tokens, call_during_try=True)
```
ptmcg: What happens if `e2` already has a parse action? I think this will replace it, which is undesirable. Add a test where `expr` is `pyparsing.common.integer`, which has a str->integer conversion parse action, and verify that after this fix that the conversion parse action still runs correctly. (If it fails, you may be able to resolve by calling `add_parse_action` instead of `set_parse_action`.)

tests/test_issue_633.py
```diff
@@ -0,0 +1,27 @@
+import pyparsing as pp
```
ptmcg: Please move these tests to tests/test_unit.py, following the pattern in the testRepeater* methods.

Be sure to add tests where expr has a parse action (suggestion: define `first` as `ppc.integer`)

pyparsing/helpers.py
fcardosojr18: Thanks I appreciate the feedback and will take a look at this in a few hours!

ptmcg:
Rebase your branch on master, I fixed the mypy-check (or rather mypy fixed their handling of lambdas)

fcardosojr18:
Rebased on latest master and addressed the review feedback:
- moved the regression coverage into tests/test_unit.py following the existing testRepeater pattern
- added coverage using pyparsing.common.integer to verify existing parse actions are preserved
- updated the implementation accordingly

Tests pass locally.

pyparsing/helpers.py
```diff
@@ -156,19 +156,20 @@ def match_previous_expr(expr: ParserElement) -> ParserElement:
...
-            theseTokens = _flatten(t.as_list())
-            if theseTokens != matchTokens:
-                raise ParseException(
-                    s, l, f"Expected {matchTokens}, found{theseTokens}"
-                )
+    def copy_token_to_repeater(s, l, t):
+        match_tokens[:] = _flatten(t.as_list())
 
-        rep.set_parse_action(must_match_these_tokens, call_during_try=True)
+    def must_match_these_tokens(s, l, t):
```
ptmcg, line 164: These updated changes are far more extensive than I had asked for. I asked you to simply change:

```python
e2.set_parse_action(must_match_these_tokens, call_during_try=True)
```

to 

```python
e2.add_parse_action(must_match_these_tokens, call_during_try=True)
```

tests/test_unit.py
```diff
@@ -2422,6 +2422,36 @@ def testRepeater5(self):
 
         self.assertParseResultsEquals(result, expected_list=expected)
 
+    def testRepeaterRecursiveWhitespace(self):
```
ptmcg, line 2425: The AI appears to have followed the pattern in the given issue, using scan_string. This makes for a messy test. Pyparsing.testing contains some useful unit test methods, see how this is done in this rewrite of testReperaterRecursiveWhitespace:

```python
    def testRepeaterRecursiveWhitespace(self):
        """test match_previous_expr with recursive whitespace"""

        if ParserElement._packratEnabled or ParserElement._left_recursion_enabled:
            print("skipping this test, not compatible with memoization")
            return

        COLON = pp.Suppress(":").set_name(":")  # for better-looking error messages
        first = pp.Char(pp.nums)
        second = pp.match_previous_expr(first)
        expr = first + COLON + second
        expr.leave_whitespace(recursive=True)

        tests = [
            ("1:1", True),
            ("1: 1", False),  # add a failed match based on whitespace
            ("1:2", False),
            ("1:a", False),
        ]

        for tst, expected in tests:
            if expected:
                expected_int_list = [t.strip() for t in tst.split(":")]
                self.assertParseAndCheckList(
                    expr, tst, expected_int_list,
                    msg=f"Failed recursive whitespace repeater test, expected pass:  {expr=} {tst=}"
                )
            else:
                with self.assertRaisesParseException(
                    msg=f"Failed recursive whitespace repeater test, expected fail:  {expr=} {tst=}"
                ):
                    expr.parse_string(tst)
```

I'm posting this comment just to pass along these notes, since you are a learning student. I can merge your tests as-is and then update to this form afterward.

fcardosojr18: Much appreciated, This is my first attempt at following another projects pattern and I hope my next attempt goes smoother.

ptmcg: Tbf, pyparsing has a number of idiosyncrasies and novel project elements, with a long, slow learning curve. Thanks for finding this bug with so simple a resolution.

fcardosojr18: I appreciate your quick responses, my first fix was due today and I will say the learning curve was far more difficult than expected, You can expect (hopefully) 5 more issues fixed over the next 4 weeks or so. I look forward to getting more comfortable with the code base and continuing to contribute.

ptmcg: If you are continuing to use AI for this work, please avail yourself of the best practices, which you can obtain by running `python -m pyparsing.ai.show_best_practices`. There are a few comments on writing unit tests, but they could stand to be improved.

ptmcg: Will you make these last changes, or should I merge what you have and then apply my stated updates?

fcardosojr18: Would you please merge, I won't have time to continue my work until the weekend

PR #638, opened 2026-06-19
Fix Located capturing leading whitespace in its location (#621)

### LINKED ISSUE #621: Some matches include erroneous leading whitespace
It seems that there is a bug which causes some matches to include erroneous leading whitespace.

pyparsing 3.1.2 and 3.3.0a1 were checked and both of them have this bug.

I am providing a minimal reproducible example below. In the example the first match works correctly and the second includes the erroneous leading whitespace.  In the example the two expressions combined via `|` are identical, but the bug happens even if the two expressions are different

```
#!/usr/bin/env python3

import pyparsing as pp

def show_match(text, loc, tokens):
	matched = text[tokens.locn_start:tokens.locn_end]
	print(f"locn_start:{tokens.locn_start}, locn_end:{tokens.locn_end}")
	print(f"Matched string '{matched}'")

abc = pp.Keyword("abc")

expr = pp.Suppress(pp.SkipTo(abc, False)) + pp.Located(abc).set_parse_action(show_match)
expr.parse_string("ignored   abc")
# Outputs
# locn_start:10, locn_end:13
# Matched string 'abc'

expr2 = pp.Suppress(pp.SkipTo(abc | abc, False)) + pp.Located(abc | abc).set_parse_action(show_match)
expr2.parse_string("ignored   abc")
# Outputs
# locn_start:7, locn_end:13
# Matched string '   abc'
```

Labib-Bin-Salam:
## Summary

Fixes #621.

`Located` decorates a match with its start/end location (`locn_start`, `locn_end`). When the wrapped expression delegates whitespace skipping to its sub-expressions — e.g. a `MatchFirst` (`a | b`) or `And` (`a + b`), whose `callPreparse` is `False` — `Located` adopts that `callPreparse=False` (in `ParseElementEnhance.__init__`) and so does **not** pre-parse the input itself. As a result `Located.parseImpl` captured `start = loc` while `loc` still pointed at the leading whitespace, and `locn_start` (and the located span) wrongly included that whitespace.

### Reproducer (from the issue)

```python
import pyparsing as pp
abc = pp.Keyword("abc")

# single token - already correct
pp.Located(abc).parse_string("   abc")        # [3, ['abc'], 6]

# MatchFirst - returned [0, ['abc'], 6] (span "   abc") before this PR
pp.Located(abc | abc).parse_string("   abc")   # now [3, ['abc'], 6]
```

This matches the diagnosis in the issue thread: *"Located is recording the start position before the inner expression applies whitespace skipping."*

## Fix

In `Located.parseImpl`, capture `start` **after** the wrapped expression's own `preParse` (`start = self.expr.preParse(instring, loc)`) instead of using the raw incoming `loc`. `preParse` honors the wrapped expression's whitespace settings, so:

- expressions that skip whitespace (the common case) now report the real start of the match, and
- expressions modified with `leave_whitespace()` still keep their leading whitespace (their `preParse` is a no-op, so `start` is unchanged).

The subsequent `self.expr._parse(..., callPreParse=False)` call is unchanged.

## Testing

- New regression test `testLocatedExprLeadingWhitespace` covering a single token, `MatchFirst`, `And`, and a `leave_whitespace()` expression. It fails on `master` (`locn_start == 0`) and passes with the fix, across all parametrized parsing configurations (packrat / left-recursion / etc.).
- Full unit suite passes: `tests/test_unit.py` (1832 passed, 6 skipped), plus `tests/test_simple_unit.py` and `tests/test_util.py` (80 passed).
- `mypy --show-error-codes --warn-unused-ignores pyparsing` — clean.
- Added a `CHANGES` entry under *Version 3.3.3 - in development*.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -19,7 +19,7 @@ jobs:
         dry-run: false
         language: python
     - name: Upload Crash
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       if: failure() && steps.build.outcome == 'success'
       with:
         name: artifacts
--- a/CHANGES
+++ b/CHANGES
@@ -33,6 +33,14 @@ Version 3.3.3 - in development
   `leave_whitespace(recursive=True)`. Issue #633, reported by Ahmed Mohamed Abdelaal,
   PR submitted by fcardosojr18 et AI.
 
+- Fixed `Located` reporting a `locn_start` that included leading whitespace when the
+  wrapped expression delegates whitespace skipping to its sub-expressions (such as
+  `MatchFirst` or `And`, whose `callPreparse` is `False`). `Located` now skips the
+  leading whitespace that the wrapped expression would skip before recording the start
+  location, so the located span no longer captures preceding whitespace. Expressions
+  modified with `leave_whitespace()` still keep their leading whitespace as before.
+  Issue #621, reported by MeanSquaredError.
+
 - Update type checking to use `mypy 1.20.0`.
 
 - 5-8% performance boost optimizing `ParseResults.copy()`. Inspired by similar changes
--- a/CHANGES
+++ b/CHANGES
@@ -34,11 +34,9 @@ Version 3.3.3 - in development
   PR submitted by fcardosojr18 et AI.
 
 - Fixed `Located` reporting a `locn_start` that included leading whitespace when the
-  wrapped expression delegates whitespace skipping to its sub-expressions (such as
-  `MatchFirst` or `And`, whose `callPreparse` is `False`). `Located` now skips the
-  leading whitespace that the wrapped expression would skip before recording the start
-  location, so the located span no longer captures preceding whitespace. Expressions
-  modified with `leave_whitespace()` still keep their leading whitespace as before.
+  wrapped expression delegates whitespace skipping to sub-expressions (such as
+  `MatchFirst` or `And`). The leading whitespace is now skipped before the start
+  location is recorded, while `leave_whitespace()` expressions keep their whitespace.
   Issue #621, reported by MeanSquaredError.
 
 - Update type checking to use `mypy 1.20.0`.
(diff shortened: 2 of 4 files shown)


ptmcg:
Looks promising. I might pare back some of these comments. Comments by their very nature will eventually diverge from the code, and the longer they are, the sooner they diverge. But they definitely capture concepts, and those are worth keeping.

I suspect some of the existing unit tests might need updating too, but I can take care of that.

Labib-Bin-Salam:
Thanks for the quick look!

- Pared back the comments in `Located.parseImpl`, the regression test, and the CHANGES entry — kept the concepts and the issue reference but trimmed the prose, since you're right that long comments drift from the code over time.
- On existing tests: the full `tests/test_unit.py` stayed green with the fix (the existing `Located` tests match `;ID PARI12345678`, with no leading whitespace before the match, so they're unaffected). Happy to leave any further test/doctest tidying to you.
- The red **Fuzzing** check is unrelated to this change: GitHub now auto-fails any workflow referencing the deprecated `actions/upload-artifact@v3`, which `.github/workflows/cifuzz.yml` still used, so the job was failing on every PR regardless of its contents. I added a small separate commit bumping it to `@v4` (a drop-in here — one artifact, uploaded only on failure) to get the check green. Glad to split that into its own PR if you'd rather keep this one focused.

Labib-Bin-Salam:
One heads-up: now that the PR touches `cifuzz.yml`, GitHub has gated the check runs behind maintainer approval (the "workflows awaiting approval" banner), so the unit tests and CIFuzz won't run until you approve them. If you'd rather keep this PR strictly to the `Located` fix, just say so and I'll drop the workflow commit and send the `upload-artifact` v3→v4 bump as its own PR.

ptmcg:
Yeeps! No unit tests failed? That is very sad to hear. I'll run your tests against the existing code to make sure they fail properly, and then add your change.

Thanks for updating cifuzz.yml, I don't have a problem including that with this other work.

Looking to rebase and merge some time today.

PR #640, opened 2026-06-24
fix: store full non-ParseResults/non-list parse action return as named result

### LINKED ISSUE #401: Inconsistent ParseResults from parse actions that return other types
The [explanation of parse action handlers](https://github.com/pyparsing/pyparsing/issues/309#issuecomment-940124991) in #309 gives three cases for the return type:
 - None: use the existing results object, potentially modified by the action
 - ParseResults: replaces the existing results object
 - Anything else: replaces the existing results object, wrapped in a ParseResults

In the "anything else" category, many types work as described.  Notice in each case that there is a single element in the tokens list, which is identical to results["letters"]:
```
>>> Empty()("name").set_parse_action(lambda: "123").parse_string("")
ParseResults(['123'], {'name': '123'})
>>> Empty()("name").set_parse_action(lambda: {1,2,3}).parse_string("")
ParseResults([{1, 2, 3}], {'name': {1, 2, 3}})
```
If the action returns a tuple, the token list is correct, but the names dict is wrong:
```
>>> Empty()("name").set_parse_action(lambda: (1,2,3)).parse_string("")
ParseResults([(1, 2, 3)], {'name': 1})
  should be: [(1, 2, 3)], {'name': (1, 2, 3)}
```
If the action returns a list, both the tokens list and names dict are wrong:
```
>>> Empty()("name").set_parse_action(lambda: [1,2,3]).parse_string("")
ParseResults([1, 2, 3], {'name': 1})
  should be: [[1, 2, 3]], {'name': [1, 2, 3]}
```
You can even get some very unexpected behavior from other types:
```
>>> Empty()("name").set_parse_action(lambda: collections.defaultdict(str)).parse_string("")
ParseResults([defaultdict(<class 'str'>, {0: ''})], {'name': ''})
  should be: [defaultdict(<class 'str'>, {})], {'name': defaultdict(<class 'str'>, {})}
```

gaoflow:
When a parse action returns a non-ParseResults, non-list value (e.g. a
tuple, set, or defaultdict), the named result was being set to only the
first element instead of the full return value. This happened because
the value was passed directly to ParseResults(), which indexes into it
with toklist[0] for the named result.

The fix wraps non-list, non-ParseResults values in [tokens] so that
toklist[0] returns the full value. Null values (empty tuple) are not
wrapped because ParseResults._null_values expects them directly. Lists
preserve their existing behavior of spreading as the token list.

Fixes #401.

This pull request was prepared with the assistance of AI, under my
direction and review.

### DIFF
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -975,13 +975,41 @@ class ParserElement(ABC):
                             raise exc from parse_action_exc
 
                         if tokens is not None and tokens is not ret_tokens:
-                            ret_tokens = ParseResults(
-                                tokens,
-                                self.resultsName,
-                                aslist=self.saveAsList
-                                and isinstance(tokens, (ParseResults, list)),
-                                modal=self.modalResults,
-                            )
+                            if isinstance(tokens, ParseResults):
+                                ret_tokens = ParseResults(
+                                    tokens,
+                                    self.resultsName,
+                                    aslist=self.saveAsList,
+                                    modal=self.modalResults,
+                                )
+                            elif isinstance(tokens, list):
+                                # Lists spread as token list (existing behavior).
+                                ret_tokens = ParseResults(
+                                    tokens,
+                                    self.resultsName,
+                                    aslist=self.saveAsList,
+                                    modal=self.modalResults,
+                                )
+                            else:
+                                # Non-ParseResults, non-list return value:
+                                # wrap so the named result stores the full
+                                # value (fixes issue #401). Null values
+                                # (empty tuple etc.) are not wrapped so
+                                # they clear the result as before.
+                                if isinstance(tokens, tuple) and len(tokens) == 0:
+                                    ret_tokens = ParseResults(
+                                        tokens,
+                                        self.resultsName,
+                                        aslist=False,
+                                        modal=self.modalResults,
+                                    )
+                                else:
+                                    ret_tokens = ParseResults(
+                                        [tokens],
+                                        self.resultsName,
+                                        aslist=False,
+                                        modal=self.modalResults,
+                                    )
                 except Exception as err:
                     # print "Exception raised in user parse action:", err
                     if self.debugActions.debug_fail:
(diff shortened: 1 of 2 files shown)


pyparsing/core.py
```diff
@@ -975,13 +975,41 @@ def _parseNoCache(
...
+                                    modal=self.modalResults,
+                                )
+                            elif isinstance(tokens, list):
+                                # Lists spread as token list (existing behavior).
+                                ret_tokens = ParseResults(
+                                    tokens,
+                                    self.resultsName,
+                                    aslist=self.saveAsList,
+                                    modal=self.modalResults,
+                                )
```
ptmcg: Could be merged into a single branch with the condition `if isinstance(tokens, (ParseResults, list)):` - fix in 2 places reduces code twice as much.

pyparsing/core.py
```diff
@@ -975,13 +975,41 @@ def _parseNoCache(
...
+                            else:
+                                # Non-ParseResults, non-list return value:
+                                # wrap so the named result stores the full
+                                # value (fixes issue #401). Null values
+                                # (empty tuple etc.) are not wrapped so
+                                # they clear the result as before.
+                                if isinstance(tokens, tuple) and len(tokens) == 0:
```
ptmcg: Change if condition to just `if tokens == ():`

binggao1230:
Applied both suggestions in `8726a50`.

The two duplicated blocks now use the compact branch:

- `ParseResults`, `list`, and `()` are passed through directly
- any other parse-action return value is wrapped as `[tokens]` so named results store the full value

Local verification:

```text
PYTHONPATH=$PWD PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest python -m pytest tests/test_unit.py::Test02_WithoutPackrat::testParseResultsNamedResultWithEmptyString tests/test_unit.py::Test02_WithoutPackrat::testParseResultsWithNamedTuple -q
git diff --check
```

ptmcg:
I'd like to add some unit tests for this. In addition, I will probably hold this back until the 3.4.0 release, since this is a potential breaking change for existing code that makes use of the current behavior.

binggao1230:
Tests added in 62d0825: `testParseActionReturningNonListType` (tuple, set, dict, defaultdict, str, int), `testParseActionReturningEmptySequence` (`()` and `[]` still clear the results name), and `testParseActionReturningList`. On master the tuple and defaultdict subtests fail; the rest pass either way.

A correction to my own description, though — "Fixes #401" overstates it. The issue also covers a returned list:

```python
Empty()("name").set_parse_action(lambda: [1, 2, 3]).parse_string("")
# ParseResults([1, 2, 3], {'name': 1});  issue expects [[1, 2, 3]], {'name': [1, 2, 3]}
```

That is unchanged here — wrapping lists too would break considerably more existing code than the tuple case does. The third test pins the current behaviour so the gap is at least visible. Happy to extend this if you'd rather have the whole issue in one go.

Agreed on holding for 3.4.0; I've left the changelog alone for that reason.

PR #642, opened 2026-06-24
fix: ParseResults __add__ and __radd__ arithmetic contract violations

gaoflow:
## Description

Two related arithmetic contract violations in `ParseResults.__add__` and `__radd__`:

**`__add__` — AttributeError instead of TypeError:** `ParseResults(["a"]) + 1` crashes with `AttributeError` (trying to access `other._tokdict` on an `int`). Per Python's data model, `__add__` should return `NotImplemented` for unrecognized operand types, letting the interpreter raise `TypeError`.

**`__radd__` — RecursionError:** `1 + ParseResults(["a"])` triggers infinite recursion. Python tries `int.__add__(1, ParseResults)` → `NotImplemented` → tries `ParseResults.__radd__(1)` → `return other + self` → `int.__add__(1, ParseResults)` → infinite loop until stack overflow.

**The fix:**
- `__add__`: guard with `isinstance(other, ParseResults)`, return `NotImplemented` otherwise
- `__radd__`: return `NotImplemented` instead of `other + self`

The `0 + ParseResults` special case (support for Python's `sum()` builtin) is preserved.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -512,6 +512,8 @@ class ParseResults:
             return ""
 
     def __add__(self, other: ParseResults) -> ParseResults:
+        if not isinstance(other, ParseResults):
+            return NotImplemented
         ret = self.copy()
         ret += other
         return ret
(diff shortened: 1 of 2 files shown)


ptmcg:
This looks good. I have a very busy rest of the week, so I'll look at packaging up PRs this weekend.

ptmcg:
Good changes, and the tests are nice.

PR #643, opened 2026-06-30
fix: deepcopy results names referencing nested ParseResults

sarathfrancis90:
I ran into this while deep-copying a parse tree so I could edit the copy and leave the original alone. `deepcopy()` copies the token list but leaves the results-name dict pointing at the original tokens, so a named token that is also a list item ends up inconsistent: reaching it by name returns the original object, reaching it by index returns the copy. Mutating either side then leaks through the other by name.

Small repro:

```python
import pyparsing as pp

grp = pp.Group(pp.Word(pp.nums)("a") + pp.Word(pp.nums)("b"))("grp")
r = (grp + pp.Word(pp.alphas)("w")).parse_string("1 2 xyz")
c = r.deepcopy()

c["grp"] is c[0]        # False -> same token, but two different objects in the copy
r["grp"]["a"] = "999"
c.as_dict()             # {'grp': {'a': '999', 'b': '2'}, 'w': 'xyz'}  -> the "deep" copy changed
```

The list item gets deep-copied but the name still resolves to the original, so `c.as_dict()` (which goes through the names) reflects edits made to the original.

The fix rebuilds `_tokdict` in `deepcopy()` so names reference the deep-copied tokens. I keep an id-map of the copied list items so a name that aliases a list item keeps aliasing the same copied object, and a name that doesn't is deep-copied on its own. After the fix `c["grp"] is c[0]` and the copy is fully independent of the original.

Added a test for the named-token case; `pytest tests examples/tiny/tests` and `mypy --show-error-codes --warn-unused-ignores pyparsing` both pass.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -50,6 +50,13 @@ Version 3.3.3 - in development
   location is recorded, while `leave_whitespace()` expressions keep their whitespace.
   Issue #621, reported by MeanSquaredError.
 
+- Fixed `ParseResults.deepcopy()` not deep-copying results names that referenced
+  nested `ParseResults`. Only the token list was copied, so a named token still
+  pointed at the original; changing the original (or the copy) was visible through
+  the other when the token was reached by name, and the same token could differ
+  when reached by index versus by name. Results names now reference the
+  deep-copied tokens. PR submitted by Sarath Francis.
+
 - Update type checking to use `mypy 1.20.0`.
 
 - 5-8% performance boost optimizing `ParseResults.copy()`. Inspired by similar changes
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -683,12 +683,16 @@ class ParseResults:
         .. versionadded:: 3.1.0
         """
         ret = self.copy()
+        # map id() of each copied token to its copy, so that results names
+        # referencing items in the token list stay linked to the copies (and
+        # decoupled from the original)
+        memo: dict[int, Any] = {}
         # replace values with copies if they are of known mutable types
         for i, obj in enumerate(self._toklist):
             if isinstance(obj, ParseResults):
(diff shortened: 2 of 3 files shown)


ptmcg:
Nice catch, and good work overall.

I'm going to take a closer look on the use of object ids as keys in the memo dict - object ids are notorious for aliasing to other objects since they can be reused when an object is deleted and a new object created. Since object deletion can occur as part of normal garbage collection/reference counting, this can happen without doing an overt 'del' command.

I'm holding at least one other PR until the 3.4 release, but I don't see any reason this couldn't go into the next 3.3.x release.

ptmcg:
After looking a little deeper into your changes, it looks like we were calling `deepcopy()` twice for embedded ParseResults: once when copying the `_toklist`, then again in `copy_value`. (Actually, the second `deepcopy()` was never called, since for ParseResults, the id was always in `memo`.) I removed the `deepcopy()` from `copy_value`, and this did in fact continue to pass your test case. Then, looking at `copy_value`, it was now pretty much the same as calling `memo.get()` with `id(value)` as the key and `value` as the default if the key is not found. So I could get rid of `copy_value` altogether.

With these changes, the block of code that builds `ret._tokdict` boils down to:

```python
        # rebuild the results-name dict so that named results point at the
        # deep-copied tokens, instead of remaining linked to the original
        ret._tokdict = {
            name: [
                _ParseResultsWithOffset(memo.get(id(value), value), offset)
                for value, offset in occurrences
            ]
            for name, occurrences in self._tokdict.items()
        }
```

I also removed the `if self._tokdict` check, the dict comprehension will generate an empty dict if `self._tokdict` is empty. No need for `copy_value`, it is largely replaced by `memo.get,` which is also much much faster.

PR #644, opened 2026-07-02
Escape control characters in set_debug() debug output

### LINKED ISSUE #496: Misleading debug text when encountering `\r`
Having a `\r` in a string which is being parsed resets the output string in the debug output, overwriting what was already there.

The output is correct, just the explanatory text

`\r` is imprtant as a standalone character, as I need to be able to accept it as a line terminator.

```python
from pyparsing import (
    Opt,
    ParserElement, 
    Regex
)

if __name__ == "__main__":

    ParserElement.set_default_whitespace_chars(" \t")
    debug = True

    line_term = (("\r" + Opt("\n")) | "\n").set_debug(flag=debug).set_name("line_term")
    comment = (Regex("#.*(?=(\r\n?)|\n)") + line_term).set_debug(flag=debug).set_name("comment")
    string = (Regex("[a-z0-9]+") + Opt(line_term)).set_debug(flag=debug).set_name("string")
    value = (string | comment).set_debug(flag=debug).set_name("value")
    file = (value[...] + line_term[...]).set_debug(flag=debug)

    s="""#multi word comment \nval val2 \r val3\nval4  \t\n\n\r"""
    print(f"{file.parse_string(s, parse_all=True)=}")
```

results in (in part):

```
#... more stuff before
Match line_term at loc 20(1,21)
  #multi word comment 
                      ^
Matched line_term -> ['\n']
Matched comment -> ['#multi word comment ', '\n']
Matched value -> ['#multi word comment ', '\n']
Match value at loc 21(2,1)
 val3
  ^
Match string at loc 21(2,1)
 val3
  ^
Match line_term at loc 24(2,4)
 val3
     ^
Match line_term failed, ParseException raised: Expected '\r', found 'val2'  (at char 25), (line:2, col:5)
Matched string -> ['val']
Matched value -> ['val']
Match value at loc 24(2,4)
 val3
     ^
Match string at loc 25(2,5)
 val3
      ^
Match line_term at loc 29(2,9)
 val3
          ^
Matched line_term -> ['\r']
Matched string -> ['val2', '\r']
Matched value -> ['val2', '\r']
#... more stuff after
```

apoorva-01:
Fixes #496. A `\r` in the parsed line got printed straight into the debug output, so the terminal ran it and the cursor jumped back to column 0, scrambling the "Match ... at loc" text. Parsing was fine, just the debug print.

Now control chars in that line are shown escaped (`\r` prints as `\r`), and the caret is nudged over so it still points at the match.

Your call on one thing: I escape all C0 controls + DEL, not just `\r`, so backspace and stray ANSI escapes can't wreck the output either. Only side effect is a tab now shows as `\t`. Happy to narrow it to `\r` only if you'd rather.

Test in `tests/test_unit.py` parses `"abc\rdef"` and checks the escaped output; fails without the fix.

closes #496

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,12 @@ Version 3.3.3 - in development
 ------------------------------
 - Adding support for Python 3.15.
 
+- Fixed debug output corruption when a parsed line contains a carriage return or
+  other control character. `set_debug()` printed the source line verbatim, so a
+  stray `\r` returned the terminal cursor to column 0 and overwrote the "Match ...
+  at loc" text. Control characters in the debug line are now shown escaped, and the
+  caret stays aligned with the match location. Issue #496, reported by Matthew Rowles.
+
 - Fixed bugs in `ParseResults.__add__()` and `ParseResults.__radd__()`, where
   erroneous `AttributeError` or `RecursionError` could be raised when adding
   `ParseResults` objects to non-`ParseResults` objects, instead of expected
--- a/pyparsing/core.py
+++ b/pyparsing/core.py
@@ -354,15 +354,27 @@ def condition_as_parse_action(
     return pa
 
 
+# control characters escaped so they can't corrupt the printed debug line
+# (for example a stray '\r' would otherwise return the cursor to column 0)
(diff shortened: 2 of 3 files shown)


ptmcg:
Looks good, thanks

PR #645, opened 2026-07-17
Give ParseResults.copy() its own results-name bookkeeping

chuenchen309:
`copy()` copies the token list, but `{**self._tokdict}` copies only the outer dict — the per-name lists of `_ParseResultsWithOffset` are the same list objects as the original's:

```python
r = expr.parse_string("1 2")
c = r.copy()
c._tokdict["b"] is r._tokdict["b"]     # True
```

`insert()` and `__delitem__()` renumber those offsets by assigning into the list in place (`occurrences[k] = _ParseResultsWithOffset(value, position + 1)`), so a mutation through the copy writes into the original's bookkeeping.

### Reproducing

```python
import pyparsing as pp
expr = pp.Word(pp.nums) + pp.Word(pp.nums)("b")

# control: no copy involved
r = expr.parse_string("1 2")
del r[0]
r.get_name()          # 'b'

# same r, same del -- but a copy was mutated in between
r = expr.parse_string("1 2")
c = r.copy()
c.insert(0, "X")
del r[0]
r.get_name()          # None
```

Observed on the untouched original: `'b'`'s offset silently moves `1 → 2` while `r._toklist` still has length 2, so the offset is out of range for its own token list. `get_name()`'s `offset <= 0` test then reads the corrupted value.

`insert(0, locn)` and `pop(0)` are the idioms in `insert()`'s and `pop()`'s own doctests, and `copy()` is on the path of `__add__` (`ret = self.copy(); ret += other`), so this is reachable without touching internals.

### Why the occurrence lists are structure, not content

`copy()`'s docstring says *"ParseResults items contained within the source are shared with the copy"* — and `testParseResultsCopy` pins that with `assertTrue(r2[1] is result[1])`. That sharing is of the contained items. `_toklist` is already copied, so the offset table is part of the structure being copied, not part of the shared content.

`deepcopy()` is the sibling that gets this right — it rebuilds `_tokdict` wholesale with fresh lists, which is why it's immune:

```python
d = r.deepcopy()
d._tokdict["b"] is r._tokdict["b"]     # False
```

Two methods maintaining the same bookkeeping, disagreeing. This is the same seam as #642/#643.

### On cost

`copy()` isn't called during parsing — I instrumented `ParseResults.copy` and ran a 50-item `DelimitedList` grammar with two results names per item: **0 calls**. So this doesn't touch parse-time performance. The occurrence lists are almost always a single element, and grammars with no results names get an empty `_tokdict` and pay nothing.

I did not run `black` over the files: it wants to reformat ~396 lines of pre-existing code in `results.py`/`tests/test_unit.py` on the untouched `master` (probably a black version difference), and none of them are mine. My lines are ≤120 and `black --diff` doesn't touch them.

### Testing

- New `testParseResultsCopyResultsNamesAreIndependent` fails on `master` and passes with the change. It asserts both halves: the occurrence lists are no longer shared, *and* the original's `get_name()` survives a mutation of the copy.
- Full suite: **1856 passed, 6 skipped, 1960 subtests** (1850 + the new test across the 6 packrat configurations). No regressions.
- CHANGES entry added.

Happy to adjust the framing or split the CHANGES entry if you'd rather.

Disclosure: found and fixed with the help of an AI coding assistant — I see the CHANGES file already uses "et AI" for this, so I've followed that convention in the entry. The repro, the call-count instrumentation, and the suite runs are ones I did myself.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -35,6 +35,12 @@ Version 3.3.3 - in development
   at loc" text. Control characters in the debug line are now shown escaped, and the
   caret stays aligned with the match location. Issue #496, reported by Matthew Rowles.
 
+- Fixed `ParseResults.copy()` sharing its results-name bookkeeping with the original.
+  The copy got its own token list, but the per-name lists of matched occurrences were
+  shared, and `insert()`/`pop()`/`del` renumber those offsets in place - so mutating a
+  copy silently corrupted the original's results names, and `get_name()` on the
+  original could return `None`. PR submitted by Andrew Chen et AI.
+
 - Fixed bugs in `ParseResults.__add__()` and `ParseResults.__radd__()`, where
   erroneous `AttributeError` or `RecursionError` could be raised when adding
   `ParseResults` objects to non-`ParseResults` objects, instead of expected
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -669,7 +669,9 @@ class ParseResults:
         """
         ret: ParseResults = object.__new__(ParseResults)
         ret._toklist = self._toklist[:]
-        ret._tokdict = {**self._tokdict}
+        # copy the occurrence lists too: insert/pop/del renumber the offsets
+        # in place, so sharing the lists would let a copy renumber the original
+        ret._tokdict = {k: v[:] for k, v in self._tokdict.items()}
         ret._parent = self._parent
         ret._all_names = {*self._all_names}
         ret._name = self._name
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -283,12 +283,15 @@ class ParseResults:
         # get removed indices
         removed = list(range(*i.indices(mylen)))
         removed.reverse()
-        # fixup indices in token dictionary
-        for occurrences in self._tokdict.values():
+        # fixup indices in token dictionary; copy() shares these lists, so
+        # renumber a private copy rather than the list itself
+        for name, occurrences in self._tokdict.items():
+            occurrences = occurrences[:]
             for j in removed:
                 for k, (value, position) in enumerate(occurrences):
                     if position > j:
                         occurrences[k] = _ParseResultsWithOffset(value, position - 1)
+            self._tokdict[name] = occurrences
 
     def __contains__(self, k) -> bool:
         return k in self._tokdict
@@ -436,11 +439,14 @@ class ParseResults:
 
         """
         self._toklist.insert(index, ins_string)
-        # fixup indices in token dictionary
-        for occurrences in self._tokdict.values():
+        # fixup indices in token dictionary; copy() shares these lists, so
+        # renumber a private copy rather than the list itself
+        for name, occurrences in self._tokdict.items():
+            occurrences = occurrences[:]
             for k, (value, position) in enumerate(occurrences):
                 if position > index:
                     occurrences[k] = _ParseResultsWithOffset(value, position + 1)
+            self._tokdict[name] = occurrences
 
     def append(self, item):
         """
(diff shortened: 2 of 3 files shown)


ptmcg:
:eyes:

ptmcg:
> copy() isn't called during parsing

Please re-confirm this with packrat parsing and left-recursion enabled.

chuenchen309:
You're right, and my "0 calls" was mis-scoped — it was measured on the **default recursive-descent** parser only. With memoization on, `copy()` **is** on the parse path:

- **Packrat**: `_parseCache` copies the result both when it stores it (`cache.set(..., value[1].copy(), ...)`) and again on every cache hit (`value[1].copy()`). On an arithmetic grammar with two results names per term I measured **~38 `copy()` calls per parse**.
- **Left recursion**: the recursion memo returns `prev_result.copy()` when it reuses an expansion — **~9 calls per parse** on `expr <<= (expr + "+" + num) | num`.

So the cost question is real. What the change actually adds, measured (2M-iter microbench):

- **No results names (empty `_tokdict`)** — the common case — is unchanged, within noise. The dict comprehension over an empty dict costs nothing.
- **With results names**, `copy()` goes ~380 ns → ~710 ns for a 3-name result (**+330 ns**), scaling with the number of names (not occurrences). Full-parse deltas on my machine are too noisy to quote a stable percentage, but for a packrat grammar where every element carries a results name it's a measurable per-parse increase.

For context: packrat's `copy()` already slices the whole token list and rebuilds the names set on every hit/miss, so copying the (usually single-element) occurrence lists is the same order of work — but it isn't free, and you were right that I stated the original claim too broadly.

If keeping `copy()` cheap on the packrat/LR hot path matters more than the localized fix, the alternative is to leave `copy()` shallow and instead have `insert()`/`pop()`/`__delitem__()` rebuild the affected occurrence list when they renumber (rebind a fresh list into `_tokdict` rather than assigning into the shared list in place). That moves the cost off the parse path onto the far rarer mutation path. I haven't implemented that yet — happy to switch to it if you'd prefer; otherwise I'll leave the fix in `copy()` as-is.

Either way I can fold these packrat/LR call counts into the PR description so the tradeoff is documented.

ptmcg:
Yes, please keep the parse processing path as optimal as possible. The changes will be a bit more complex in the mutation path, but we'll live with it.

chuenchen309:
Done — `9c2bd72` moves the copying off `copy()` and onto the two methods that renumber.

Only `__delitem__` and `insert` rewrite the recorded offsets; `__setitem__` and `__iadd__` already build new lists rather than mutating in place. So `copy()` can go back to `{**self._tokdict}` and share the occurrence lists, and those two renumber a private copy instead. Nothing else ever mutates a shared list, so a copy can't renumber the original.

200k iterations each, best of five, unpatched → copying in `copy()` → copying when renumbering:

| | unpatched | previous | now |
|---|---|---|---|
| `copy()`, 3 results names | 311 ns | 637 ns | 323 ns |
| `copy()`, no results names | 277 ns | 365 ns | 276 ns |
| `insert()` | 1192 ns | 1173 ns | 1548 ns |
| `del r[0]` | 1626 ns | 1581 ns | 2067 ns |

**One correction to what I told you earlier.** I said the no-results-names case — the common one — was "unchanged, within noise". That was wrong: the dict comprehension cost about 90 ns there too, and it reproduces across repeated runs rather than washing out. That's the row that should have mattered most for your decision, and I gave you the wrong number for it. The version now in the branch is back to baseline on both rows.

The regression test no longer asserts on `_tokdict` identity, since sharing is now the intended state; it checks the behaviour in both directions instead (mutate the copy, the original's name survives; mutate the original, the copy's name survives). It fails six ways on `181babf` — once per parser configuration — and passes here.

`tests/` is 2013 passed / 33 skipped, unchanged from the branch point. `test_examples.py::test_range_check` fails at both, and `test_diagram.py` needs `railroad`, which I don't have installed.

PR #646, opened 2026-07-19
Make Word(max=...) match up to max regardless of whitespace in the charset

chuenchen309:
### The divergence

`Word(..., max=n)` gives opposite results depending on whether the character set happens to contain whitespace:

```python
import pyparsing as pp

pp.Word(pp.nums, max=3).parse_string("0123456")        # ['012']  — matches up to max, leaves the rest
pp.Word(pp.nums + " ", max=3).parse_string("0123456")  # raises ParseException
```

The only difference is a user-invisible internal path choice (`core.py`): a `Word` whose charset has no whitespace uses the regex fast path, which matches up to `max` and leaves the remainder for the next parser; a charset *with* whitespace falls back to the char-by-char `Word.parseImpl`, which instead **raises** once the run of body chars exceeds `max`. So the same public API + same `max` behaves differently based on the charset contents.

The lenient behavior is the one the default (regex) path produces, the one `exact=` already produces in both paths, and the one asserted by the active `tests/test_unit.py::testWordMinMaxExactArgs`. The strict-throw branch in the fallback looks like a leftover: `set(triple_quoted)`-style, it's the one place the two paths disagree on `max`.

### The change

Drop the stale strict-throw branch so the fallback matches the regex path. This breaks zero existing tests; I've added a regression test asserting both charsets yield `["012", "3456"]`, and the full suite passes (1856 passed, 6 skipped).

### Direction is your call

I want to flag honestly that this is a behavior choice, not a clear-cut fix. `CHANGES` for v1.3.1 (2005) documents the **strict** semantics ("`Word(nums, max=3)` ... will raise an exception" on `'0123456'`) as an intentional feature — but that predates the regex fast path and `testWordMinMaxExactArgs`, which encode **lenient**. I went with lenient because it (a) makes the fallback consistent with the path most inputs actually take, (b) matches `exact=`'s existing behavior, and (c) breaks no current tests. The alternative — keep strict and add a bounded-vs-greedy guard to the regex path (plus updating `testWordMinMaxExactArgs`) — is larger and would change currently-passing behavior. Happy to switch to that direction, or close this, if you'd prefer strict.

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the divergence, verified it, wrote the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,13 @@ Version 3.3.3 - in development
 ------------------------------
 - Adding support for Python 3.15.
 
+- Fixed `pyparsing_common.as_datetime` raising `Invalid date/time: microsecond
+  must be in 0..999999` for valid ISO-8601 timestamps whose fractional seconds
+  round up to a full second (e.g. `2021-06-15T12:30:59.9999995`, common in
+  nanosecond-precision timestamps). The rounded microseconds are now added via
+  `timedelta` so the value carries into the next second instead of overflowing
+  the `datetime` microsecond argument. PR submitted by Andrew Chen et AI.
+
 - Fixed debug output corruption when a parsed line contains a carriage return or
   other control character. `set_debug()` printed the source line verbatim, so a
   stray `\r` returned the terminal cursor to column 0 and overwrote the "Match ...
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,11 @@ Version 3.3.3 - in development
 ------------------------------
 - Adding support for Python 3.15.
 
+- Fixed `ParseResults.__iadd__` (``+=``) leaking an `AttributeError` about a
+  private attribute when the right-hand operand is not a `ParseResults` (e.g.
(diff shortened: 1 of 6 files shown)


ptmcg:
I agree. While adding " " to the characters in a Word is not recommended, the behavior should not be so different. I'll add this to my weekend todo list.

ptmcg:
The code change looks good. Making some minor changes in your test method:
- remove the "Regression: ..." part of the docstring that was inserted by the AI. Yes, this is a "regression" in that you are fixing a bug, and thus code is different from before
- Change your validation code to use one of the supplied helper methods that packages up the call to parse_string, the validation of the result, and failure message

ptmcg:
I made a mess of rebasing, extracted the changes and committed to master directly. I'll add attribution in the CHANGES doc.

ptmcg:
Attributed in CHANGES doc

PR #647, opened 2026-07-19
Carry a rounded-up fractional second in as_datetime instead of raising Invalid date/time

chuenchen309:
### Problem

`pyparsing_common.as_datetime` raises `ParseException: Invalid date/time: microsecond must be in 0..999999` for valid ISO-8601 timestamps whose fractional-seconds field rounds up to a full second:

```pycon
>>> from pyparsing import pyparsing_common as ppc
>>> expr = ppc.iso8601_datetime().add_parse_action(ppc.as_datetime)
>>> expr.parse_string("2021-06-15T12:30:59.9999995")
pyparsing.exceptions.ParseException: Invalid date/time: microsecond must be in 0..999999
>>> expr.parse_string("2021-01-01T00:00:00.999999999")
pyparsing.exceptions.ParseException: Invalid date/time: microsecond must be in 0..999999
```

Every input field here is valid — the year/month/day/hour/minute/second are all in range. The `microsecond` that trips the `datetime` constructor is an internally-computed value, so the "Invalid date/time" message points at a field the user never supplied. The `iso8601_datetime` regex accepts arbitrary fractional digits (`\d\d(\.\d*)?`), so sub-microsecond precision (Go `RFC3339Nano`, Kubernetes/Prometheus/systemd logs, DB timestamps) reaches this path in normal use.

### Root cause

`pyparsing/common.py`, in `as_datetime`:

```python
round((second % 1) * 1_000_000),   # can be exactly 1_000_000
```

`round(...)` can produce `1_000_000`, but `datetime`'s microsecond argument must be `0..999999`. This line was added in #639 to fix fractional-second mis-scaling; the `round()` closed the floor error but opened a boundary overflow at fractions ≥ ~0.9999995 s.

### Fix

Add the rounded microseconds via `timedelta`, so a value that rounds up to a full second carries into the next second instead of overflowing:

```python
return datetime(year, month, day, hour, minute, int(second)) + timedelta(
    microseconds=round((second % 1) * 1_000_000)
)
```

### Verification

- Before the fix: `"2021-06-15T12:30:59.9999995"` and `"2021-01-01T00:00:00.999999999"` raise; a new subtest in `testCommonExpressions` fails.
- After the fix: they return `datetime(2021, 6, 15, 12, 31, 0)` and `datetime(2021, 1, 1, 0, 0, 1)`; the subtest passes.
- Unchanged: `0.45 → 450000`, `0.999999 → 999999` (existing assertions still pass), and a literal `"60"` leap second still raises `second must be in 0..59` because `int(second)` is untouched.
- Full `tests/test_unit.py`: 1850 passed, 6 skipped, 0 failed.

The carry (vs. clamping to 999999) matches stdlib `timedelta` rounding; if you'd prefer a different boundary policy, happy to adjust.

---

*Disclosure: this PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro, wrote the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.*

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,13 @@ Version 3.3.3 - in development
 ------------------------------
 - Adding support for Python 3.15.
 
+- Fixed `pyparsing_common.as_datetime` raising `Invalid date/time: microsecond
+  must be in 0..999999` for valid ISO-8601 timestamps whose fractional seconds
+  round up to a full second (e.g. `2021-06-15T12:30:59.9999995`, common in
+  nanosecond-precision timestamps). The rounded microseconds are now added via
+  `timedelta` so the value carries into the next second instead of overflowing
+  the `datetime` microsecond argument. PR submitted by Andrew Chen et AI.
+
 - Fixed debug output corruption when a parsed line contains a carriage return or
   other control character. `set_debug()` printed the source line verbatim, so a
   stray `\r` returned the terminal cursor to column 0 and overwrote the "Match ...
--- a/pyparsing/common.py
+++ b/pyparsing/common.py
@@ -1,7 +1,7 @@
 # common.py
 from .core import *
 from .helpers import DelimitedList, any_open_tag, any_close_tag
-from datetime import datetime
+from datetime import datetime, timedelta
 import sys
 
 PY_310_OR_LATER = sys.version_info >= (3, 10)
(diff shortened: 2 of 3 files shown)


ptmcg:
Nice fix. I'm not sure I'd agree that this is "common" with nanosecond timestamps, but if you do something a million times a day, something that can only happen once every other 1/1000000th of a second will still happen every other day.

PR #650, opened 2026-07-30
perf: stop rebuilding a results-name occurrence list on every append

zhangj23:
Setting a results name that already exists appends to its occurrence list by rebuilding the whole list, so producing n occurrences of one name costs O(n^2). A grammar using a `name*` list-collector over a long input pays this directly.

This appends in place instead, guarded by a small `_own_keys` set recording which names' lists the instance is sole owner of. Every path that shares a list hands ownership back, so an in-place append can never reach another instance's list.

## Numbers

`time.process_time`, alternating base and patched subprocesses, min of 2 runs, appending n occurrences of one results name:

| n    | master    | this branch | speedup | master growth | branch growth |
|-----:|----------:|------------:|--------:|--------------:|--------------:|
| 1000 |   1.10 ms |     0.53 ms |    2.1x |             - |             - |
| 2000 |   3.97 ms |     0.74 ms |    5.3x |         x3.60 |         x1.40 |
| 4000 |  19.89 ms |     1.57 ms |   12.6x |         x5.01 |         x2.12 |
| 8000 |  86.34 ms |     3.15 ms |   27.4x |         x4.34 |         x2.00 |

`master` grows roughly 4.3x per doubling against this branch's 2x, so the quadratic term is gone.

This is narrow, and it is a trade rather than a free win. It only helps a results name that accumulates many occurrences, and it costs a little on the common case.

An ordinary parse making a single named assignment is about 6 percent slower. Median of 6 alternating runs of 30,000 parses of `Word(alphas)("w")`:

```
master:      59.66 ms
this branch: 63.19 ms   (0.944x)
```

That is the cost of maintaining the `_own_keys` ownership set on a path that never benefits from it, since a name with one occurrence has nothing to amortise.

So: roughly 6 percent on single-occurrence names, in exchange for removing the quadratic on names that accumulate many, where it is 27x at n=8000 and still growing. Whether that trade is worth taking is your call, and I would understand closing this if the common case matters more.

If you would prefer, I am happy to try a version that only starts tracking ownership once a name reaches a second occurrence, which should leave the single-assignment path untouched. I did not want to push a redesign without asking first.

## Behaviour

The ownership set is cleared in `copy()`, `__delitem__`, `clear()`, `__setstate__`, `__deepcopy__` and `__getstate__`.

`__getstate__` is the one that matters and it was missing in my first draft. `ParseResults` defines no `__copy__` and no `__reduce__`, so `copy.copy` routes through `__getstate__` then `__setstate__`, and `__getstate__` returns `self._tokdict.copy()`, which is shallow. Without clearing ownership there, the source keeps handing out lists the copy now shares, and appending to the source also grew the copy:

```python
r = grammar.parse_string("abc")
c = copy.copy(r)
for i in range(4):
    r["w"] = "src%d" % i

# master:  c["w"] == 'abc',  len 1
# unfixed: c["w"] == 'src3', len 5
```

Differential testing against `master`: 45 comparisons over 5 grammars, crossing parse, repeated append, `copy.copy`, `copy.deepcopy`, pickle round-trip, `ParseResults.copy()`, append-to-source-after-copy, and `del`, comparing `as_list`, `as_dict` and `dump` output. 0 mismatches.

## Checks

`pytest` gives 2056 passed, 27 skipped, exit code 0, matching `master` exactly. The tox `mypy-check` gate passes with no issues in 17 source files.

`black --check` reports a diff on `pyparsing/results.py`, but the pristine file fails identically (it wants a blank line removed after the `.util` import), so that is pre-existing rather than introduced here, and `black` is not in the tox envlist.

### DIFF
--- a/pyparsing/results.py
+++ b/pyparsing/results.py
@@ -86,6 +86,7 @@ class ParseResults:
     _modal: bool
     _toklist: list[Any]
     _tokdict: dict[str, Any]
+    _own_keys: set[Any] | None
 
     __slots__ = (
         "_name",
@@ -94,6 +95,7 @@ class ParseResults:
         "_modal",
         "_toklist",
         "_tokdict",
+        "_own_keys",
     )
 
     class List(list):
@@ -186,6 +188,7 @@ class ParseResults:
         else:
             self._toklist = [toklist]
         self._tokdict = {}
+        self._own_keys = None
         return self
 
     # Performance tuning: we construct a *lot* of these, so keep this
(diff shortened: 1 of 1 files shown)


ptmcg:
Optimizing the addition of another value for a given results name in a ParseResults seems like a worthy effort, but I don't see the point in the `_own_keys` set, nor in the added function. Why not just get the value if it exists and append to it?

```python
@@ -248,16 +248,23 @@
         return ParseResults([v.result for v in self._tokdict[i]])
 
     def __setitem__(self, k, v, isinstance=isinstance):
+        no_value = object()
         if isinstance(v, _ParseResultsWithOffset):
-            self._tokdict[k] = self._tokdict.get(k, list()) + [v]
+            cur_tokdict_value = self._tokdict.get(k, no_value)
+            if cur_tokdict_value is no_value:
+                self._tokdict[k] = [v]
+            else:
+                cur_tokdict_value.append(v)
             sub = v.result
         elif isinstance(k, (int, slice)):
             self._toklist[k] = v
             sub = v
         else:
-            self._tokdict[k] = self._tokdict.get(k, []) + [
-                _ParseResultsWithOffset(v, 0)
-            ]
+            cur_tokdict_value = self._tokdict.get(k, no_value)
+            if cur_tokdict_value is no_value:
+                self._tokdict[k] = [_ParseResultsWithOffset(v, 0)]
+            else:
+                cur_tokdict_value.append(_ParseResultsWithOffset(v, 0))
             sub = v
         if isinstance(sub, ParseResults):
             sub._parent = self
```

How does this stack up against your performance testing?  Are there unit tests to show the need for `_own_keys`?

ptmcg:
My tests show a 29% perf improvement with my simplified change. If I don't hear back from you this week, I'll go ahead with that.

FYI - the original code:
```python
self._tokdict[k] = self._tokdict.get(k, list()) + [v]
```
dates back at least to 2012.

ptmcg:
I committed the simplified change that I posted earlier, with attribution to your PR. Thanks for submitting!

PR #653, opened 2026-08-15
Fix match_previous_expr to match nested and repeated expressions (issue #560)

### LINKED ISSUE #560: match_previous_expr does not handle nested expressions
See [this SO post](https://stackoverflow.com/questions/78574784/how-to-make-sure-element-has-parsing-precedence-before-other-elements-in-pyparsi), should be able to use match_previous_expr to enforce matching of opening and closing tags. Does not work for nested tags.

rajeshkanade:
Fixes #560.

### Cause

`copy_token_to_repeater` adds a new `must_match_these_tokens` guard to the repeated expression on
every match of the leading expression, and never removes one. After N matches the expression carries
N guards, each closed over a different set of tokens, and all of them run on every attempt to match
the repeat — so at most one can succeed.

That breaks more than nesting:

| input | expected | before |
|---|---|---|
| `{a}xx{/a}` | match | match |
| `{a}{b}xx{/b}{/a}` | match | fails |
| two sequential pairs, `a:a b:b` under `OneOrMore` | match | fails |
| reparsing valid input after one failed parse | match | fails |

The guard list also grows without bound for the lifetime of the expression.

### Change

Keep the pending matches on a stack of `(location, tokens)` and register the guard once at
construction, so an inner match pairs with the nearest enclosing one. Entries recorded at or after
the current location are dropped on the way in, so matches left behind by a branch that was
backtracked out of can never be paired — that keeps the stack bounded (it returns to empty after
every parse; measured at 2000 sequential pairs and at 200 levels of nesting).

A memoizing parser can run the guard twice for the same match, so the guard returns early if it is
re-entered for the location and tokens it just matched. Without that, `match_previous_expr` under
`enable_left_recursion()` regressed relative to master.

The `ParseException` message is unchanged, byte for byte, in case anyone matches on it.

### Tests

Two tests added next to the existing `testRepeater*` tests, both failing before this change and
passing after:

- `testRepeaterNested` — three nesting depths, plus crossed and mismatched close tags that must
  still be rejected.
- `testRepeaterReusedParser` — repeated matches in one parse, and that a mismatch does not stop the
  same expression from matching valid input afterwards.

Full suite: 2097 passed before, 2109 after (the 12 are the two new tests across the test classes),
27 skipped either way. `mypy --show-error-codes --warn-unused-ignores pyparsing` is clean and the
change is black-clean.

Checked against `enable_packrat()` and `enable_left_recursion()` as well: both fix the repeat case
and neither regresses against master. Also checked `scan_string`/`search_string`, lookahead contexts
(`NotAny`, `FollowedBy`, `Opt`), and two independent `match_previous_expr` instances in one grammar.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -94,6 +94,14 @@ Version 3.3.3 - in development
   `leave_whitespace(recursive=True)`. Issue #633, reported by Ahmed Mohamed Abdelaal,
   PR submitted by fcardosojr18 et AI.
 
+- Fixed `match_previous_expr()` failing to match nested or repeated expressions, so that
+  it can now be used for matching nested open/close tags. A new guard parse action was
+  added on every match of the repeated expression and never removed, so all of them ran
+  on each attempt to match and only one could succeed; a second pair anywhere in the
+  input would fail, and after a failed parse the expression would keep rejecting valid
+  input. The pending matches are now kept on a stack, so an inner match pairs with the
+  nearest enclosing one. Issue #560, PR submitted by Rajesh Kanade et AI.
+
 - Fixed `Located` reporting a `locn_start` that included leading whitespace when the
   wrapped expression delegates whitespace skipping to sub-expressions (such as
   `MatchFirst` or `And`). The leading whitespace is now skipped before the start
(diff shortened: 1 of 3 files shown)


ptmcg:
I'm a little anxious about the `match_stack` being shared this way. I had a similar design in an early version of `IndentedBlock`, and the stack was not managed well (I don't remember whether by me or by the client code), so I eventually converted the helper method to a class so that the indent stack would be managed in the class instance. With your code, I tried creating multiple `match_previous_expr `expressions to see if they would step on each other's matches, but the only way I could force it was to make the input string invalid.  Still :eyes:...

ptmcg:
I'm going to go ahead and merge this as-is, and make this next release a bumped minor version to reflect that may be behavior changes.

Thanks for submitting.

Issue #608, opened 2025-06-26
Documentation part Deux


ferdnyc:
#607 is merged (yay!), containing my first-pass documentation changes. Those were the things that were fairly easy, low-effort tweaks to implement, and that I was confident enough in submitting that I felt OK doing them "cold".

But there's more that could be done. There's more I'd _like_ to do. But it's stuff that now involves either enough work, or enough prior discussion, that I'd prefer to get consensus on my plans for it **before** I put in the work. So, here are some documentation proposals / topics for discussion, things that if we agree they sound useful, I'd be interested in working on or helping with next.

### The examples

A while back I submitted #595 which corrected the example output for the `ParseException` example (and _just_ the `ParseException` example) to reflect code changes since it was written.

@ptmcg pointed out, quite correctly, that there were plenty more where that came from. (Meaning, examples that are no longer up-to-date with current code.) I'd like to address those, but I'd like to address them in a manner that won't immediately go back to bit-rotting out of sync with the code.

So, what I'd propose is to turn the example code in the docs into doctests. (Ideally **all** of it, or as close as we can come without making things harder for ourselves.) Even the ones that don't actually output anything, can be configured as no-output doctests, and running the tests will at least verify that those examples _execute_ properly.

I started experimentally doing that with a few of them, and I already found one code error. This example won't run, because the first code line has an extraneous close paren at the end:

https://github.com/pyparsing/pyparsing/blob/4752d13cd9086d8ef40641ea016baa55d1db46fe/docs/whats_new_in_3_1.rst?plain=1#L32-L40

Doctests can catch small typos like that, because when the examples are executed and that line throws a `SyntaxError`, it'll get flagged. And for the examples that _do_ show output, they'll ensure that the example is kept in sync with code changes, because the doctest will start failing whenever code changes alter the output from what's expected in the example.

### The theme

When I first started using pyparsing, the documentation was presented in the Sphinx Alabaster theme, which personally I quite liked. Then at some point that changed to the Classic theme, which personally I think looks like hot garbage.

But I never really heard any reason for the change, so I'm curious why the switch was made? 

If it was about the table of contents (IIRC Alabaster didn't handle local TOCs the same way classic does by default), I believe I can configure Alabaster to have the same TOC structure Classic uses by default. If that's possible, would switching back to Alabaster be a consideration? Are there other motivating reasons beyond the TOC (assuming that was even part of the reason to begin with)?

### The organization

[...]

Issue #509, opened 2023-08-28
RecursionError with Forward syntax


RobinetDenisAcsone:
Follow up of https://github.com/pyparsing/pyparsing/issues/489#issuecomment-1691432208
Summary: Having a syntax with nested function can cause a RecursionError based on the number of nesting

@ptmcg As you asked, here is the syntax code: https://github.com/Arelle/Arelle/blob/master/arelle/formula/XPathParser.py#L799

A bit of context of use:
Arelle is a software that handle a specification called XBRL, which allow reporting data based on a taxonomy (which define what kind of data) created by an authority.

In one of the latest version of a taxonomy ([solvency 2.8](https://www.eiopa.europa.eu/tools-and-data/supervisory-reporting-dpm-and-xbrl_en)) there are a few formulas (used to validate the data) which looked like this:
```
iaf:numeric-equal($v1,iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add(iaf:numeric-add($v2,$v3),$v4),$v5),$v6),$v7),$v8),$v9),$v10),$v11),$v12),$v13),$v14),$v15),$v16),$v17),$v18),$v19),$v20),$v21),$v22),$v23),$v24),$v25),$v26),$v27),$v28),$v29),$v30),$v31),$v32),iaf:sum($v33)))
```
(id: `s2md_BV311_1-1-1`)

Compiling it result to a RecursionError with the syntax provided above.

If you want to replicate the real use case, I can explain how, but be aware that it take a few minutes of loading to come to that point.

@ptmcg I tried to be as short as possible in the explanation, but fell free to ask me anything.
The 2 main thing to understand is:
* We can't change what is accepted in the syntax (since it's defined in a specification)
* The string to parse is also outside our control.

[...]

Issue #403, opened 2022-05-28
infix_notation running out of stack


mishamsk:
Hi,

I am building a grammar for a subset of SQL (or more specifically [SAS PROC SQL sql-expression](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.5/sqlproc/p020urejdmvi7vn1t9avbvazqapu.htm), which is an extension of ANSI SQL).

I've used infix_notation, however it generates what seems to be a very non-optimal parser. Specifically expressions like this in the infix_notation
`matchExpr = _FB(lastExpr + opExpr + lastExpr) + Group(
                        lastExpr + (opExpr + lastExpr)[1, ...]
                    )`  force parser to recurse and backtrack (due to Forward rules), which grows out of proportion for deep nested expressions. If set_debug is enabled, one can see how it forwards and then backs at least 3 times even for simplest expressions (except for a single term ones). I am not sure why the forwards are needed in the first place.

So even simple string like '`PUT(PUT(Personeelsnummer, z6.),$50.)`' would parse around a second (I am on MacBook Pro M1 Pro with 32gigs). And this string:
```
 WHEN DATEPART(W25LQE8.CMN30_PostDatum) ne DATEPART(W25LQE8.CMN30_AanmaakDatum) AND W25LQE8.CMN30_PostDatum ne .
     THEN   DATEPART(W25LQE8.CMN30_PostDatum)*60*60*24 + 60*60*10*1*1*1*1*1*1*1*1*1*1*1*1*1*1 /*Post eerdere datum dan gescand, dan 10 uur 's-ochtends op postdatum*/
    WHEN WKCAP6.SCAN_DTM_START_DTD ne . THEN WKCAP6.SCAN_DTM_START_DTD /*post op dezelfde dag binnen als gescand, dan eerste scandatum van de dag*/
    ELSE DATEPART(COALESCE(W25LQE8.CMN30_PostDatum, W25LQE8.CMN30_AanmaakDatum))*60*60*24 + 60*60*10 /*Als geen eerste scandatum van de dag, dan 10 uur s-ochtends*/
    END
```
would crash running out of stack with `RecursionError: maximum recursion depth exceeded`. Raising limit to 2000 `sys.setrecursionlimit(2000)` would allow it to parse...in 6 seconds!

So I just rebuilt infix_notation based on the original code and your own example with [numbers](https://github.com/pyparsing/pyparsing/blob/master/examples/fourFn.py) like so:

```
[...]

Issue #204, opened 2020-04-07
infixNotation, parentheses, performance


Igreh:
Hi! 
We are faced with high performance degradation issue for strings with multiple nesting levels. Just curious - is it bug or feature? Are there any workarounds?

To reproduce:
Take pyparsing/examples/simpleArith.py and add new tests:
~~~python
    "(a + b)",
    "((a + b))",
    "(((a + b)))",
    "((((a + b))))",
~~~
each new pair of parentheses slow parsing process significantly.
Anyway, thanks for such a great tool!

ptmcg:
Good catch! I added your examples and the parser does indeed slow to a crawl.

The typical solution to this problem for any recursive parser (as in all parsers defined using infixNotation) is to enable packrat parsing.  I'm surprised I haven't added that to this example.

Immediately after importing pyparsing, add this line:

    ParserElement.enablePackrat()

With this change (and bumping up the sys recursion limit to 3000), I can now parse this expression almost instantly:

        "((((((((((((((a + b))))))))))))))",

Igreh:
Works like a charm!
Thanks!

PR #655, opened 2026-08-26
Document that packrat parsing consumes additional stack depth

Gares95:
`enable_packrat()`, its entry in `docs/HowToUsePyparsing.rst`, and `pyparsing/ai/best_practices.md` all recommend enabling packrat for recursive grammars, and none of them mentions that memoizing costs stack depth. A grammar that parses input at a given nesting depth without packrat can raise `RecursionError` at that same depth with packrat enabled.

This is already your own guidance rather than anything new. In #204 you answered a nesting-performance report with "enable packrat parsing ... (and bumping up the sys recursion limit to 3000)". This PR puts that second half somewhere users will find it. It is not a bug report, and it is not about `RecursionError` on deeply nested input, which I take from #27, #403 and #509 to be working as intended.

The mechanism, for reference: `enable_packrat()` sets `ParserElement._parse = ParserElement._parseCache`, and `_parseCache` calls `_parseNoCache` on a cache miss, so each `_parse` invocation costs an extra frame.

Changes, documentation only with no behavior change:

- `enable_packrat()` docstring: a short paragraph noting the trade-off and pointing at `sys.setrecursionlimit()`.
- `docs/HowToUsePyparsing.rst`: the same note on the `enable_packrat()` entry. `best_practices.md` links readers to this page, so it seemed wrong to add the caveat in one place and not the other.
- `pyparsing/ai/best_practices.md`: one sub-bullet under the existing packrat recommendation. That file ships in the wheel and is written to be read by AI agents, which will follow an unconditional recommendation literally.
- `CHANGES`: an entry under 3.3.3.

Happy to drop any of these, or reword them, if you would rather keep a given file shorter. I also noticed the docs work under discussion in #608 and did not want to cut across it, so this is deliberately narrow.

Checks run locally on Python 3.13 in a container:

- `mypy --show-error-codes --warn-unused-ignores pyparsing` - clean, 17 source files
- `pytest tests examples/tiny/tests` - 2157 passed, 2041 subtests passed
- `sphinx-build -b doctest docs docs/_build/doctest` - 265 tests, 0 failures, and no new build warnings
- `black --check pyparsing/core.py` - unchanged


ptmcg:
I have some issues in flight on a private commit branch on this subject right now, so there are some changes already taking place in this area that will probably affect what we eventually have to say about packratting and recursion in pyparsing. Plus I will be offline for the next two weeks. So I will hang on to this work, and when I return, the other issue should have resolved and we  can revisit.

Thanks for submitting, I appreciate the research you've put into this. I just want to set your expectations that I will not be able to do anything directly with your PR for about the next 2-3 weeks.

Gares95:
Thanks, and that all makes sense. No rush at all on this one, and enjoy the time off. It is clearly better to let your in-flight work settle first, and I am happy to rebase or rework this around whatever lands.

One separate thing worth flagging before you go, since it is in the same file and it is yours to prioritise. While looking at the builtin quoted string expressions I found that moving the closing quote into the regex in dd391e8 made them backtrack exponentially on input with no closing quote. `'"' + r"\xAA" * 26`, an input of 105 characters, takes about 12 seconds in `quoted_string.parse_string`, and each extra repetition roughly doubles it. The hex escape branch can hand trailing hex digits back to the general body branch, so `\xAA` has two possible partitions and k of them have 2**k.

The important part for your peace of mind while offline: this is master only. dd391e8 is in no tag and PyPI is still on 3.3.2, so no released version is affected and nobody is exposed. It can safely wait.

I have a fix and a test ready if you want them. The escape branch takes a single hex digit and the general branch absorbs the rest, which accepts exactly the same strings with only one possible partition, so matching is linear and valid input parses slightly faster too. The existing `testReCatastrophicBacktrackingInQuotedStringParsers` does not catch it, because its cases repeat a single character and never reach the escape branch.

Happy to open that as a PR now, hold it until you are back, or drop it entirely if your private branch already covers it. Whichever is least disruptive. No need to reply before you return.

ptmcg:
Wow, that is a great catch on the quoted_string regex! I recently did some deeper study on regexes and backtracking, and thought that the issue no longer pertained. If you would submit a separate PR for these changes, I can merge that into master, to go out in the release after next. (Work in this area will not conflict with the other changes that are in progress.)

And I'm fairly certain I have a unit test specifically for this, please update it with your test string. (Check back in the git log to see how I had resolved it earlier, and if your changes end up being more optimal, that would be super - my earlier resolution of this split the parsing of the trailing quote into a separate expression, so that regex backtracking was bypassed, but adding another pyparsing expression in this way is a bit of a performance drag, on a fairly frequently used builtin.)

Thanks again,
-- Paul

Gares95:
Opened as #656.

Two things changed since my note above, and both belong here rather than only in the PR description.

It turned out to be three instances of the same thing rather than one. Moving the closing delimiters into the regexes exposed three separate pairs of overlapping body alternatives, and the hex escape I described was only the first. The others are the `\"` alternative in `python_quoted_string`, which its general escape alternative already matches since a quote is not `x`, and the triple-quoted bodies, which match both one and two quotes at a time. `"""` + `""a` * 18 is 57 characters and shows the same doubling. All three are in the PR, since they share one cause and one unreleased window.

I also need to correct something I said above. I said valid input parses slightly faster with the fix. I could not reproduce that with a specified benchmark, so I am withdrawing the speedup claim: over six representative valid strings, `quoted_string.parse_string` measures 39.9 to 50.0 usec on master against 41.9 to 42.7 with the change across repeated runs, which overlap, and raw `re.match` is 0.617 against 0.624 usec best of five. I found no repeatable valid-input difference in either, and my earlier wording overstated it.

No need to look at any of this before you are back.

PR #657, opened 2026-09-04
Retain exactly capacity items in LRUMemo

youdie006:
`LRUMemo` retains `capacity + 2` items. Its own docstring (`pyparsing/util.py:130`) says:

> A memoizing mapping that retains `capacity` deleted items ... once `capacity` items are retained,
> the least recently used item is discarded.

`pyparsing/util.py:160`:

```python
oldest_keys = list(self._memory)[: -(self._capacity + 1)]
for key_to_delete in oldest_keys:
    self._memory.pop(key_to_delete)
self._memory[key] = value
```

The slice keeps the last `capacity + 1` entries and then inserts one more, so the steady state is
`capacity + 2`. `capacity=0`, which should retain nothing, retains 2.

Setting and deleting 20 keys:

```
                    before          after
capacity=0   retained=2      retained=0
capacity=1   retained=3      retained=1
capacity=2   retained=4      retained=2
capacity=4   retained=6      retained=4
```

The user-facing knob is `ParserElement.enable_left_recursion(cache_size_limit=n)`
(`pyparsing/core.py:1177`), which constructs `LRUMemo(n)` - so asking for a 4-entry memo gives 6.

### The change

Insert first, then evict while over capacity.

I did not use the one-token fix `[: -(self._capacity - 1)]`, because at `capacity == 1` Python
evaluates `x[:-0]` as `x[:0]`, which is empty - nothing would ever be evicted and the memo would grow
without bound. The loop form also handles `capacity == 0` without a special case.

### Why nothing caught it

`Test10_WithLeftRecursionParsingBoundedMemo` (`tests/test_unit.py:11890`) sets `cache_size_limit=4`
and then asserts on the stored setting, never on what is retained:

```python
self.assertLessEqual(ParserElement.recursion_memos._capacity, 4)
self.assertGreater(ParserElement.recursion_memos._capacity * 3, 4)
```

with the comment `# it may be larger due to action handling`. `_capacity` is whatever was passed in,
so those two lines pass regardless of the eviction logic. `grep -rn "LRUMemo" tests/` finds no other
reference.

### Tests

A parametrized test in `tests/test_util.py` over `capacity` 0, 1, 2, 4, asserting both the retained
count and that the retained keys are the most recent ones. Mutation-checked in both directions:

- revert to the slice -> `assert 2 == 0`, `assert 3 == 1`
- over-correct to `> self._capacity + 1` -> `assert 1 == 0`, `assert 2 == 1`

so it pins the boundary from both sides rather than only from below.

`pytest tests/test_unit.py tests/test_simple_unit.py tests/test_util.py` gives
`2018 passed, 6 skipped, 2033 subtests passed`. That includes
`Test10_WithLeftRecursionParsingBoundedMemo`, which re-runs the whole `Test02` parsing suite under
`cache_size_limit=4` - so shrinking retention from 6 to 4 does not affect left-recursion parsing, and
the overshoot was not load-bearing.

**Observable change, stated plainly:** a bounded left-recursion memo now holds exactly
`cache_size_limit` retained entries instead of `cache_size_limit + 2`, and `LRUMemo(0)` retains
nothing. No parse results change.

`CHANGES` updated with the `et AI` annotation per `CONTRIBUTING.md`.

---

Disclosure: prepared with AI assistance. Every figure above is from a run on this branch, and the
root cause is one I traced and verified by hand.


ptmcg:
Looks good, thanks for submitting. (I'm just back from vacation, so I'm a little backed up on PRs this week, hope to work through them soon. This one all looks good though.)

Issue #482, opened 2023-05-12
NotAny(FollowedBy()) failing


genry718:
i have some input data like

`[gog1] [G1] [gog2] [gog3] [gog4] [G2] [gog5] [G3] [gog6]`

and want to find all gogs, if no G after it. so in this case i want to get `gog2` and `gog3` (and maybe `gog6`).

looks pretty simple, rigth? but i failed :(

```python
import pyparsing as pp
from pyparsing import *

def pyparsing_test():
    # this also dont helps
    # ParserElement.enable_left_recursion(force=True)

    data=""" [gog1] [G1] [gog2] [gog3] [gog4] [G2] [gog5] [G3] [gog6] """

    poi_type = Word(alphas).set_results_name('type')
    poi = Suppress('[') + poi_type + Char(nums) + Suppress(']')

    def cnd_is_type(val):
        return lambda toks: toks.type==val

    def cnd_is_not_type(val):
        return lambda toks: toks.type!=val

    poi_gog=poi('gog').add_condition(cnd_is_type('gog'))
    poi_g=poi('g').add_condition(cnd_is_type('G'))
    poi_not_g=poi('not_g').add_condition(cnd_is_not_type('G'))

    pattern = poi_gog + ~poi_g
    #WTF this finds only `gog6`, why??

    pattern = poi_gog + NotAny(FollowedBy(poi_g))
    #WTF same, only `gog6`

    pattern = poi_gog + poi_not_g.suppress()
    #WTF this works better but find only `gog2`, why not `gog3` also?

    r=pattern.search_string(data)
    print(data)
    print('='*10)
    print(r)
```

so my research shows that `NotAny(FollowedBy(poi_g))` not works at all. what im doing wrong?

ptmcg:
I saw your post on SO, and have looked into it  a bit. Part of the problem is that results names will ordinarily be treated like dict keys - the last value assigned to a key will overwrite all previous values. Setting expression names and set_debug() will print out more of the step-by-step details of how the parser is working. As a short-cut, you can add these lines before calling search_string:

```python
    pp.enable_diag(pp.Diagnostics.enable_debug_on_named_expressions)
    pp.autoname_elements()
```

Changing your last pattern to:

```python
    pattern = poi_gog + FollowedBy(Group(poi_not_g) | StringEnd())
```

Gives the gog2, gog3, gog6 result.

[...]

PR #658, opened 2026-09-05
fix: parse actions on Or/MatchFirst now receive the whitespace-skipped loc

### LINKED ISSUE #557: set_parse_action handler receives seemingly erroneous loc argument
Essentially, I am using `set_parse_action()` to determine to location of parse elements within the source string (if there's another way, I'm all ears). The parse action defined in the [documentation](https://pyparsing-docs.readthedocs.io/en/latest/pyparsing.html#pyparsing.ParserElement.set_parse_action) as having a fingerprint of:

```
Each parse action fn is a callable method with 0-3 arguments, called as fn(s, loc, toks) , fn(loc, toks) , fn(toks) , or just fn() , where:

    s = the original string being parsed
    loc = the location of the matching substring
    toks = a list of the matched tokens, packaged as a ParseResults object
```

And so I write me one to test this:

```python
def test_value_location(s, loc, toks):
    pfx = "\t\t"
    print(f"{pfx}{s=}")
    print(f"{pfx}{loc=}")
    print(f"{pfx}{toks=}")
    value1 = ''.join(toks)
    value2 = toks.asDict()['value']
    value3 = s[loc:loc+len(value1)]
    print(f"{pfx}{value1=}")
    print(f"{pfx}{value2=}")
    print(f"{pfx}{value3=}")
```

and implement a parser. Only I find that the loc that is reported is inconsistent and in one scenario IMHO broken (either a bug, or surprisingly unintuitive and hard to comprehend behaviour that warrants clear documentation - which is may even have and I've not found it).

The issue seems to be that if value follows a white space element be that Empty(*), White() or '' loc is out by 1! 

To confirm this, I wrote a test script (attached as:  [pyparsing-bug.py.zip](https://github.com/pyparsing/pyparsing/files/15454516/pyparsing-bug.py.zip))

In synopsis, it defines two test lines to parse, and 4 scenarios to test. The first line is of the form "setting = value" the second of the form "setting value", and the scenarios define an assignment operator as '=', Empty(), White() or ''. In each scenario are parse action handler prints its arguments and three versions of the value parsed (that handler is above). The first values if from toks as a list, the second from toks as a dict, the third from the source string at loc. 

I expect all the values to be identical in all successfully parsed scenarios and I expect the first line to parse only in the first scenario and the second line in the remaining three scenarios. And that is exactly what I find except that the third value (from the source string at loc) demonstrates an inconsistency. That value is at loc in the empty scenarios seems right and out by one in the non-empty scenario. 

Surely the whole point of loc, is to point to the value parsed regardless?

The full output in which you can see value3 is wrong in Line 0 scenario 1:

```
Line 0: setting_1 = value-1 # comment 1
	Scenario 0: '='
		s='setting_1 = value-1 # comment 1\n'
		loc=11
		toks=ParseResults(['value-1'], {'value': 'value-1'})
		value1='value-1'
		value2='value-1'
		value3=' value-'
		Result:
			['setting_1', '=', 'value-1', '#', ' comment 1']
			- name: 'setting_1'
			- trailing_comment: 			['#', ' comment 1']
			- value: 'value-1'

	Scenario 1: Empty
	Failed to parse: Expected {quoted string using single or double quotes | W:(-.0-9A-Za-z)}, found '='  (at char 10), (line:1, col:11)

	Scenario 2: <SP><TAB><CR><LF>
	Failed to parse: Expected {quoted string using single or double quotes | W:(-.0-9A-Za-z)}, found '='  (at char 10), (line:1, col:11)

	Scenario 3: ''
	Failed to parse: Expected {quoted string using single or double quotes | W:(-.0-9A-Za-z)}, found '='  (at char 10), (line:1, col:11)

Line 1: setting_2 value-2 # comment 2
	Scenario 0: '='
	Failed to parse: Expected '=', found 'value'  (at char 10), (line:1, col:11)

	Scenario 1: Empty
		s='setting_2 value-2 # comment 2\n'
		loc=10
		toks=ParseResults(['value-2'], {'value': 'value-2'})
		value1='value-2'
		value2='value-2'
		value3='value-2'
		Result:
			['setting_2', 'value-2', '#', ' comment 2']
			- name: 'setting_2'
			- trailing_comment: 			['#', ' comment 2']
			- value: 'value-2'

	Scenario 2: <SP><TAB><CR><LF>
		s='setting_2 value-2 # comment 2\n'
		loc=10
		toks=ParseResults(['value-2'], {'value': 'value-2'})
		value1='value-2'
		value2='value-2'
		value3='value-2'
		Result:
			['setting_2', ' ', 'value-2', '#', ' comment 2']
			- name: 'setting_2'
			- trailing_comment: 			['#', ' comment 2']
			- value: 'value-2'

	Scenario 3: ''
		s='setting_2 value-2 # comment 2\n'
		loc=10
		toks=ParseResults(['value-2'], {'value': 'value-2'})
		value1='value-2'
		value2='value-2'
		value3='value-2'
		Result:
			['setting_2', 'value-2', '#', ' comment 2']
			- name: 'setting_2'
			- trailing_comment: 			['#', ' comment 2']
			- value: 'value-2'
```

dualfroz:
## Problem

Fixes: https://github.com/pyparsing/pyparsing/issues/557
Base: upstream `master` @ `efd56db4e59f36b1673ce0eb0823e3afaa9d1201`
Branch: `dualfroz/fix-or-matchfirst-parseaction-loc`

## Problem (#557)

A parse action attached directly to an `Or` (`^`) or `MatchFirst` (`|`)
expression receives a `loc` argument that still points at the leading
whitespace preceding the matched tokens, instead of at the first character
of the match. The same construct built with `And` (`+`) is correct.

The original issue report (a config-line parser with optional separators)
is a more elaborate manifestation of this same root cause; this PR isolates
and fixes the underlying mechanism rather than the specific repro in the
issue.

## Root cause

`ParserElement._parseNoCache` (`pyparsing/core.py`, ~line 928) decides
whether to whitespace-skip *before* recording `tokens_start` -- the `loc`
that gets handed to a parse action -- based on the element's own
`self.callPreparse` flag:

```python
if callPreParse and self.callPreparse:
    pre_loc = self.preParse(instring, loc)
else:
    pre_loc = loc
tokens_start = pre_loc
```

`ParseExpression.__init__` (`pyparsing/core.py:4325`, the shared base class
for `And`, `Or`, `MatchFirst`, `Each`) sets `self.callPreparse = False` by
default.

`And.__init__` (`pyparsing/core.py:4547`) explicitly corrects this back to
`self.callPreparse = True` after computing its own `skipWhitespace` /
`whiteChars` from its first sub-expression, so a parse action attached
directly to an `And` sees the correct, whitespace-skipped `loc`.

`Or.__init__` (`pyparsing/core.py:4699-4707`, pre-fix) and
`MatchFirst.__init__` (`pyparsing/core.py:4859-4867`, pre-fix) compute
`self.skipWhitespace = all(e.skipWhitespace for e in self.exprs)` but never
touch `self.callPreparse`, so it stays `False`. As a result, `pre_loc ==
loc` (unskipped) for `Or`/`MatchFirst`, and any parse action attached
directly to one of them sees the raw, un-skipped `loc` -- even though
`Or.parseImpl` (line ~4726) and each `MatchFirst` alternative internally
skip whitespace correctly when matching, so the *tokens* and the returned
end-of-match `loc` are fine; only the `loc` argument seen by the
directly-attached parse action is wrong.

## Fix

`pyparsing/core.py`: in both `Or.__init__` and `MatchFirst.__init__`, set

```python
self.callPreparse = all(e.callPreparse for e in self.exprs)
```

alongside the existing `self.skipWhitespace = all(e.skipWhitespace for e in
self.exprs)` line (only in the `if self.exprs:` branch, mirroring where
`skipWhitespace` is already set).

This condition is deliberately the same one `Or.parseImpl` already uses
internally (`if all(e.callPreparse for e in self.exprs): loc =
self.preParse(instring, loc)`) to decide whether to pre-skip whitespace
before trying alternatives. Reusing it means:

- When every alternative would itself pre-skip whitespace, `Or`/
  `MatchFirst` now also pre-skip once at the top of `_parseNoCache`, so
  `tokens_start` (and hence the parse action's `loc`) is correct. The extra
  `preParse` call already inside `Or.parseImpl`/each `MatchFirst`
  alternative's own `_parse()` becomes a harmless no-op (it starts from an
  already-skipped position using the exact same `whiteChars`/
  `skipWhitespace` state), so the position ultimately used to try each
  alternative -- and therefore *which* alternative wins -- is unchanged.
- When at least one alternative would *not* pre-skip (e.g. a `White()`
  alternative, or `AtStringStart`/`AtLineStart`, which intentionally
  disable `callPreparse`), `self.callPreparse` stays `False`, exactly as
  before, so matching semantics for those alternatives are untouched.

`Or.streamline()`/`MatchFirst.streamline()` were deliberately left alone:
they flatten nested same-type expressions by splicing in the *same* child
expression objects (see `ParseExpression.streamline`, core.py:4383), so
`all(e.callPreparse for e in self.exprs)` evaluates to the same boolean
before and after flattening (AND is associative); no re-derivation is
needed there. `Or.parseImpl`/`MatchFirst.parseImpl` were also left
untouched, per scope.

Only `pyparsing/core.py` (the two `__init__` methods) and
`tests/test_unit.py` (the new regression test) were changed.

## Empirical repro (before fix), on this branch's base commit

Standalone script (not part of the commit, kept locally for reference):

```
input string: '   abc' (token starts at index 3)

And ('+')                        parsed=['abc']      loc=3    expected=3    [OK]
Or ('^')                         parsed=['abc']      loc=0    expected=3    [BUG]
MatchFirst ('|')                 parsed=['abc']      loc=0    expected=3    [BUG]

CONFIRMED: And receives correct loc; Or/MatchFirst do not.
```

## After fix

```
input string: '   abc' (token starts at index 3)

And ('+')                        parsed=['abc']      loc=3    expected=3    [OK]
Or ('^')                         parsed=['abc']      loc=3    expected=3    [OK]
MatchFirst ('|')                 parsed=['abc']      loc=3    expected=3    [OK]

All three receive correct loc; bug not reproduced on this checkout.
```

## New regression test: `testParseActionLocOnOrAndMatchFirst` in `tests/test_unit.py`

Added right after `testParseActionRunsInNotAny` (issue #482), following the
existing `testXxx` / `# see Issue #NNN` style in that section. It builds
`word ^ Literal("xyz")` and `word | Literal("xyz")` over an input with
leading whitespace, attaches a parse action directly to each, and asserts
the recorded `loc` equals the post-whitespace token start.

Because pyparsing's `Test0N_...` classes replay every test method under
several parsing modes (no packrat / packrat / bounded packrat / unbounded
packrat / left-recursion / left-recursion+bounded memo), this one method
runs 6 times.

## Unrelated, pre-existing environment gap (not a regression)

`tests/test_diagram.py` and one case in `tests/test_examples.py`
(`test_range_check`) fail to collect/run because the optional `railroad`
package (`pyparsing[diagrams]` extra) is not installed in this venv. This
is reproducible identically on an unmodified checkout of the same base
commit and is unrelated to this change; the task's specified gate
(`tests/test_unit.py`) does not depend on it.


ptmcg:
Thanks for this PR and it makes sense. I may hold off on it until 3.4.0, since it has a good chance of being a breaking change, since it modifies the location passed to parse actions.

dualfroz:
sounds good, thanks

PR #659, opened 2026-09-06
Fix reset_pyparsing_context parse action leaks

### LINKED ISSUE #171: reset_pyparsing_context does not remove any added parse actions
If a global like quotedString has had a parse action (like removeQuotes) added to it, `reset_pyparsing_context` does not remove that parse action afterward.

This may not have a viable solution, so at least might merit some apologetic documentation.

cucuwang:
Fixes #171.

## Summary

`reset_pyparsing_context` now snapshots and restores built-in expressions' parse actions and `callDuringTry` state. Parse-action lists are restored in place, with regression coverage for nested contexts.

## Tests

- `python -m pytest tests examples/tiny/tests -q` (`2130 passed, 27 skipped, 2041 subtests passed`)
- `mypy --show-error-codes --warn-unused-ignores pyparsing`
- `black --fast --check pyparsing/testing.py`

## AI assistance

Codex was used to implement, test, and review this change. The required `et AI` attribution is included in `CHANGES`.


ptmcg:
Looks good, adding to my PR to-merge list (which is a bit long since I've been out of town, so ...)

PR #661, opened 2026-09-07
Fix White respecting default whitespace policy

### LINKED ISSUE #359: pp.White skips preceding whitespace characters even if characters removed from default (from issue #350 resolution)
From revision 1ccf846 onward (the fix to #350), if the default whitespace characters are removed, for example, if there are no default whitespace characters, `pp.White(ws)` appears to skip over characters in " \t\n\r" to consume `ws` even if those characters are not in `ws` and not in the default whitespace characters:

```python
import pyparsing as pp
pp.ParserElement.setDefaultWhitespaceChars("")

("a" + pp.White(" ") + "c").parseString("a\n c")  # Matches, should not.
#[Out]# ParseResults(['a', ' ', 'c'], {})

("a" + pp.White(" ") + "c").parseString("a\t c")  # Matches, should not.
#[Out]# ParseResults(['a', '        ', 'c'], {})

("a" + pp.White("\n") + "c").parseString("a\nc")  # Matches, should.
#[Out]# ParseResults(['a', '\n', 'c'], {})

("a" + pp.White("\n") + "c").parseString("a \nc")  # Matches, should not.
#[Out]# ParseResults(['a', '\n', 'c'], {})

("a" + pp.White("\n") + "c").parseString("a \n c")  # Does not match, should not.

("a" + pp.White("\n") + "c").parseString("a\n c")  # Does not match, should not.
```

emecii:
Fixes #359.

`White` previously built its pre-parse whitespace set from every whitespace character it knows about. This overrode a caller's `ParserElement.set_default_whitespace_chars()` policy: after setting it to `""`, `White(" ")` still skipped a leading newline before matching a space.

Build the set from the configured default whitespace characters instead, excluding only the characters `White` is meant to consume. The added regression test confirms that a space still matches while a preceding newline is rejected when default whitespace is disabled.

Validation:

- `pytest tests/test_unit.py -k 'White or CombineWithResultsNames' -q` — 72 passed
- `pytest tests/test_unit.py -q` — 1,946 passed
- `git diff --check`

The changelog follows the project's AI-assisted attribution convention.


ptmcg:
I am back from vacation, and there are a few priority PRs ahead in the queue. I may hold off on this until 3.4.0 since it changes the behavior of White.

PR #663, opened 2026-09-09
Fix unit CI jobs selecting an empty tox environment

### LINKED ISSUE #662: `Unit tests` CI jobs have run no tests since 6622151 — ci.yml's `TOXENV=py` selects an env with no commands
Since 6622151 ("Update tox.ini to add doctest testing for all Python versions 3.10 and later") the `Unit tests` jobs have not run the test suite. `.github/workflows/ci.yml` sets `toxenv: [py]` (line 24) and `TOXENV: ${{ matrix.toxenv || 'py' }}` (line 33), so eight of the nine jobs run `tox -e py`. That commit made every dep, extra and command in `[testenv]` factor-conditional (`unit:` / `doctest:`), and `py` carries neither factor, so it now resolves to an empty env: tox builds the sdist, installs it, prints `congratulations :)` and exits. `mypy-check` is the only leg still running a check.

This is CI-only. A bare `tox` uses `env_list` and still runs `py39-unit`, `py310-unit` and the rest; it is the `TOXENV` variable in ci.yml, which overrides `env_list`, that selects the empty env. Those two lines in ci.yml are unchanged — they are the same in 6622151's parent.

At master (efd56db):

```console
$ tox c -e py -k commands deps extras
[testenv:py]
commands =
deps =
extras =
```

In CI it looks like this — [run 32099081148](https://github.com/pyparsing/pyparsing/actions/runs/32099081148) (the most recent master run), job `Unit tests (ubuntu-latest, py, 3.12)`, the tail of its `tox` output from the sdist install onward:

```
py: install_package> python -I -m pip install --force-reinstall --no-deps .../pyparsing-3.3.3.tar.gz
py: freeze> python -m pip freeze --all
py: pip==26.2.1,pyparsing @ file:///.../pyparsing-3.3.3.tar.gz#sha256=ed824222...
.pkg: _exit> python .../pyproject_api/_backend.py True flit_core.buildapi
  py: OK (5.41 seconds)
  congratulations :) (5.45 seconds)
```

The job clock dates it. On the last master run before that commit ([17521582908](https://github.com/pyparsing/pyparsing/actions/runs/17521582908), 23092da) the 3.12 leg took 101 s; on the first run containing it ([17521713797](https://github.com/pyparsing/pyparsing/actions/runs/17521713797), fad1505), 15 s. On the ten most recent master runs it is 14–19 s. ecf9357 ("Re-enable running matplotlib unit tests", 2026-08-09) has not run in CI either.

Two lines restore it:

```diff
         os: ["ubuntu-latest"]
-        toxenv: [py]
+        toxenv: [py-unit]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
@@
     env:
-      TOXENV: ${{ matrix.toxenv || 'py' }}
+      TOXENV: ${{ matrix.toxenv || 'py-unit' }}
```

I ran that on a fork of master at efd56db — the same nine jobs both times, the only other change being the two repro branch names added to ci.yml's push trigger so the workflow fires there:

- your config: [run A](https://github.com/glaziermag/pyparsing/actions/runs/34270753319) — nine green; every `py` leg ends `py: OK` in 4.4–6.6 s, with no `collected N items` line and no pytest output at all.
- with the two lines: [run B](https://github.com/glaziermag/pyparsing/actions/runs/34270761334) — all eight test legs report `collected 2157 items`; 2157 passed on 3.9–3.14 and macOS, 2130 passed / 27 skipped on pypy-3.9. Nine green.

So restoring the tests does not turn CI red, and the `include:` legs (macOS 3.14, pypy-3.9) pick the env up through the `|| 'py-unit'` default. The change deliberately leaves the `doctest` envs off — they need sphinx and myst-parser, which the workflow does not install, and I did not test them.

Filing this rather than a PR: it is your CI config, and which factors to run is your call.

Disclosure: I ran the `tox c` above at efd56db, read the linked job log, and ran both fork runs; everything is linked, so none of it has to be taken on trust. Written with AI assistance (Claude Code) — noted here in the spirit of CONTRIBUTING's "et AI" convention, which covers CHANGES entries rather than issues.

Neallin-917:
## Summary

Closes #662. Applies the minimal fix diagnosed and proposed by @glaziermag in that issue:

- Select `py-unit` in the unit-test matrix.
- Use `py-unit` as the fallback for the macOS and PyPy include entries.
- Keep the explicit `mypy-check` job and existing doctest scope unchanged.

The `unit:` factor gates the dependencies, diagrams extra and pytest command in `tox.ini`. Selecting `py` currently builds and installs the package but runs no tests. Both CI selections need the unit factor to restore coverage for all unit-test jobs.

## Validation

- Before: `uvx tox c -e py,py-unit -k commands` confirms an empty command list for `py` and `pytest tests examples/tiny/tests` for `py-unit`.
- Before: an actual `uvx tox -e py` run in an isolated work directory completed successfully in 4.18 seconds without executing tests.
- After: `TOXENV=py-unit uvx tox` on macOS / Python 3.14.0 ran the packaged project with its diagrams extra and matplotlib: **2157 passed in 29.17 seconds**, tox exit 0. The complete cross-platform matrix is left to upstream CI; this is the local result, not a claim of running every matrix entry.
- `git diff --check` passes. No changes to the library, tox commands, Python matrix, dependencies or doctest jobs.

AI assistance was used to implement and validate this change. `CHANGES` includes the required `submitted by Neal Lin et AI` attribution and credits the issue author's diagnosis and proposed fix.


ptmcg:
I ran the pipeline, I can see the tests running now - looks good, thanks!

